PR #2, opened 2026-01-23
Initial implementation

easybe:
Moved commits from `main` to branch for review.


smart_system_local/devices/dynamic/device_builder.py
```diff
@@ -0,0 +1,219 @@
...
+
+        model_loader = await get_model_loader()
+        model_definition = model_loader.get_model(model_number)
+        if not model_definition:
+            raise ValueError(f"Unknown model number: {model_number}")
+
+        return cls(id=device_id, raw=device_data, model_definition=model_definition)
+
+    @staticmethod
+    def discover() -> list[dict[str, Any]]:
```
broglep-work: currently hardcoded to lemonbeatd. must support providing the service as parameter (or produce discovery request for all services)

smart_system_local/devices/dynamic/device_builder.py
```diff
@@ -0,0 +1,219 @@
...
+                f"Resource {resource_name} not found in {object_name}[{instance_id}]"
+            )
+
+        if not resource.is_writable:
+            raise ValueError(
+                f"Resource {object_name}/{instance_id}/{resource_name} is not writable"
+            )
+
+        # Determine payload type from value
+        if isinstance(value, bool):
```
broglep-work, line 335: note that this does not support all cases, the type itself is not sufficient to find the bnw protocol type. also date time are not properly supported. only part support for arrays.

complete reference:

https://dev.azure.com/HQV-Gardena/SG-Gateway/_git/sg-bnw-lib?path=/bnw/protocol/src/bnw/protocol/__init__.py&version=GBmain&line=74&lineEnd=75&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

smart_system_local/devices/dynamic/device_builder.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        obj = self.get_object(object_name, str(instance_id))
+        return obj.get_value(resource_name, self.raw) if obj else None
+
+    def set_resource(
+        self,
+        object_name: str,
+        instance_id: str | int,
+        resource_name: str,
+        value: Any,
```
broglep-work, line 303: `value: Any` won't be expressive enough. the python type itself it not sufficient always to exactly define the bnw protocol type, in most cases it works, but we need a optional argument to pass a type hint. 

reference: https://dev.azure.com/HQV-Gardena/SG-Gateway/_git/sg-bnw-lib?path=/bnw/protocol/src/bnw/protocol/__init__.py&version=GBmain&line=203&lineEnd=204&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

smart_system_local/devices/base.py
```diff
@@ -0,0 +1,259 @@
...
+        elif isinstance(value, str):
+            payload = {"vs": value}
+        else:
+            raise TypeError(f"Unsupported value type: {type(value)}")
+
+        # Build the path from the enum tuple
+        path_str = "/".join(path.value)
+        
+        return {
+            "request-id": str(uuid.uuid4()),
```
broglep-work: We should use a RequestModel instead of plain dicts (we already have a response model, why no request model?)

smart_system_local/schema/device_models.yaml
broglep-work, line 1: we need some transparency (documentation, scripts, processes) on how this file was produced, otherwise it will be impossible to check if this is complete (it most likely isn't fully). E.g. opaque value type is completely missing which were heavily used by leamonbeat devices in the past. 

I got some background & idea from where this is coming from (data api), I would like to question if we can re-use this description here as both of them serve different purposes. My recommendation would be to generate such a file from the firmware description (codegen yaml files [sg-zephyr-firmware](https://dev.azure.com/HQV-Gardena/SG-Embedded/_git/sg-zephyr-firmware?path=/codegen)  and icd yaml for legacy devices [e.g. ic24](https://dev.azure.com/HQV-Gardena/SG-Embedded/_git/sg-lb-firmware?path=/application/ic24/icd_ic24.yml))

scripts/ws_listener.py
broglep-work, line 1: I'm missing an abstraction of the communication channel to the gateway (that wraps the websocket and provides one one method to get the async event stream, and another method allow to perform a request)

broglep-work (commented):
I have general doubts  about the foundation / approach. Why are we re-inventing partially everything from scratch (and doing it with only partial completeness) when we already have complete & proven foundation with bnw-lib (I'm only talking about [bnw.protcol](https://dev.azure.com/HQV-Gardena/SG-Gateway/_git/sg-bnw-lib?path=/bnw/protocol/src/bnw/protocol) and not the other parts are not well suited for this project here)? 

We also have some somewhat useful concepts on how to expose the protocol to a human (developer or end user) with the bnw-cli and bnw-lib (e.g. `bnw.transport`), we could let us inspire from it and the experience we gained with designing those interfaces

smart_system_local/devices/base.py
```diff
@@ -0,0 +1,259 @@
...
+        obj: Any = self.raw
+        for key in path_parts:
+            if not isinstance(obj, dict):
+                return None
+            obj = obj.get(key)
+        if isinstance(obj, dict):
+            return ValueField(**obj).value
+        return obj
+
+    def get_field(self, *path: str) -> ValueField | None:
```
LukasWoodtli: `Optional[ValueField]` is usually preferred instead Union (`|`)  with `None`.
Check in the whole project.

smart_system_local/devices/base.py
LukasWoodtli: Agree!

smart_system_local/devices/base.py
```diff
@@ -0,0 +1,259 @@
...
+                "request-id": "2a8166c5-d60f-4ddd-8735-29aa3661a128",
+                "op": "read",
+                "entity": {"device": "device_id", "path": "lemonbeat/0/power_timer"},
+                "metadata": {}
+            }
+        """
+        # Build the path from the enum tuple
+        path_str = "/".join(path.value)
+        
+        return {
```
LukasWoodtli: Use here also a model instead of a plain dict.

smart_system_local/schema/device_models_schema.json
LukasWoodtli, line 1: I really like the use of JSON schema and `pydantic`. It seems to be the modern way of doing this kind of stuff and familiar to a lot of developers. It's going to help when we eventually want to attract contributors from outside the company.

README.md
LukasWoodtli, line 1: We will need a more detailed README eventually. But that can follow later.

LukasWoodtli:
> I have general doubts about the foundation / approach. Why are we re-inventing partially everything from scratch (and doing it with only partial completeness) when we already have complete & proven foundation with bnw-lib (I'm only talking about [bnw.protcol](https://dev.azure.com/HQV-Gardena/SG-Gateway/_git/sg-bnw-lib?path=/bnw/protocol/src/bnw/protocol) and not the other parts are not well suited for this project here)?

I agree that we could reuse [bnw/protocol](https://dev.azure.com/HQV-Gardena/SG-Gateway/_git/sg-bnw-lib?path=/bnw/protocol/src/bnw/protocol). It's quite small and decoupled from the rest. But using modern tools like `pydantic` will be more familiar for potential future contributors. The addition of JSON schema adds some level of documentation and flexibility regarding tooling and technologies. Moreover, the protocol is simple and easily reimplemented using these modern tools.

> We also have some somewhat useful concepts on how to expose the protocol to a human (developer or end user) with the bnw-cli and bnw-lib (e.g. `bnw.transport`), we could let us inspire from it and the experience we gained with designing those interfaces

Taking inspiration from the functionality of `bnw-lib` (and `bnw-cli`) could be useful, but keep in mind that their implementations are quite complex. We don't need that level of sophistication here. We should prioritize straightforward solutions and clear, approachable code so we don't discourage future contributors.

easybe:
> We should prioritize straightforward solutions and clear, approachable code so we don't discourage future contributors.

I Very much agree. At FOSDEM we talked to the openHAB people and they stressed the point that the API should be well documented. I generally believe that code should be written in a way that it may serve as documentation. As openHAB is written in Java, they will be  reimplementing this.

scripts/ws_listener.py
kelvan: the ws scripts are just for debugging, the websocket is only a dev dependency, the idea was that the actual websocket part goes into the integration implementation, but I'm open for discussion :)

smart_system_local/devices/base.py
kelvan: afaik `| None` is the prefered way with 3.10+
But it is a matter of taste (I would prefer Optional too, but | is the more common way nowadays and saves an import)

smart_system_local/devices/dynamic/device_builder.py
kelvan: @easybe could I send both discovery requests in one message?

easybe: No, not in websocketd's current state.

broglep-work: Theoretically it would be possible as the protocols allows for it, as a message is always a list of protocol messages, you could send two request in one websocket message. Cloudadapter at least routes each request individually (when multiple are in one), but I have never tested if two request towards two different services work as expected. I guess for websocketd it is probably similar. It adds some complexity of routing and also the need to synchronize & handle some edge cases (what if one request fails and the other suceeds)

easybe: Yes, we could adapt websocketd to handle multiple messages at once. If that makes things easier I could look into that.

kelvan: I can send two separate discover messages, shouldn't be a problem

easybe:
We have decided to do this in smaller steps. First part: https://github.com/cloudless-garden/py-gardena-smart-system-local/pull/5

PR #3, opened 2026-01-23
Add LICENSE file

easybe:


rettichschnidi (commented):
LGPL ist just an addition to the GPL, will need both licenses.

https://www.gnu.org/licenses/gpl-howto.html.en#license-files

easybe:
> LGPL ist just an addition to the GPL, will need both licenses.
> 
> https://www.gnu.org/licenses/gpl-howto.html.en#license-files

Fixed. Downloaded https://www.gnu.org/licenses/lgpl+gpl-3.0.txt

rettichschnidi (approved):

PR #5, opened 2026-02-12
Add support for Water Control (18869)

### LINKED ISSUE #4: Add minimal device specs yaml
- Only fields we need
- Start with Water Control support

kelvan:
Minimal implementation only supporting Water Control

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,53 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  qa:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      pull-requests: write
+    strategy:
+      matrix:
+        python-version: ["3.12", "3.13", "3.14"]
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+
+      - name: Install dependencies
+        run: |
+          uv sync --all-groups
+          uv pip install -e .
+
+      - name: Run type checking
+        run: |
+          uv run ty check
+
+      - name: Run linting
+        run: |
+          uv run ruff check
+
+      - name: Run tests with coverage
+        run: |
+          uv run pytest --cov=gardena_smart_local_api --cov-report=term --cov-report=xml
+
+      - name: Coverage comment
+        if: github.event_name == 'pull_request'
+        uses: py-cov-action/python-coverage-comment-action@v3
+        with:
+          GITHUB_TOKEN: ${{ github.token }}
--- a/README.md
+++ b/README.md
@@ -1,0 +1,31 @@
+# Python library for GARDENA smart local API
+
+Enables controlling and monitoring GARDENA smart devices in the local network, without going through the cloud.
+
+## Supported Devices
+
+| Model | Device |
+|-------|--------|
+| 18869 | Water Control |
+
+## Installation
+
+```bash
+uv sync
+```
+
+## Linting
+
+```bash
+uv sync --all-groups
+uv run ruff check
+uv run ruff format --check
+uv run ty check
+```
+
+## Running Tests
+
+```bash
+uv sync --group test
+uv run python -m pytest
+```
--- a/gardena_smart_local_api/devices/device.py
+++ b/gardena_smart_local_api/devices/device.py
@@ -1,0 +1,263 @@
+from collections.abc import MutableMapping
+from functools import cached_property
+from typing import Any, ClassVar
+
+from pydantic import BaseModel, Field, RootModel
+
+from ..messages import EgressMessageList, Entity, Event, Request
+from ..model_loader import ModelDefinition, get_model_loader
+from ..resources import VALUE_TYPES, IpsoObject, IpsoPath, ValueField
+from ..utils import deep_merge_dict, delete_nested_key
+
+
+def _value_to_payload[T: VALUE_TYPES](value: T) -> dict[str, T]:
+    TYPE_MAP = {
+        bool: "vb",
+        int: "vi",
+        float: "vf",
+        str: "vs",
+    }
+
+    value_type = type(value)
+
+    if value_type in TYPE_MAP:
+        return {TYPE_MAP[value_type]: value}
+    elif isinstance(value, list):
+        if value and isinstance(value[0], int):
+            return {"ai": value}
+        else:
+            return {"as": value}
+    else:
+        raise TypeError(f"Unsupported value type: {value_type}")
+
+
+class Device(BaseModel):
+    id: str
+    data: dict[str, Any] = Field(repr=False)
+    model_definition: ModelDefinition = Field()
+    service: ClassVar[str] = ""
+
+    @classmethod
+    async def _from_raw(cls, device_data: dict[str, Any]) -> "Device | None":
+        device_id, device_data = next(iter(device_data.items()))
+
+        model_number: str | None = (
+            device_data.get("device", {}).get("0", {}).get("model_number", {}).get("vs")
+        )
+        if model_number is None:
+            raise ValueError("Could not extract model_number from device data")
+
+        model_loader = await get_model_loader()
+        model_definition = model_loader.get_model(model_number)
+        if model_definition is None:
+            return None
+
+        return cls(
+            id=device_id,
+            data=device_data,
+            model_definition=model_definition,
+        )
+
+    def update_data(self, event: Event) -> None:
+        if event.op == "delete":
+            delete_nested_key(self.data, event.entity.path)
+            return
+
+        if not event.payload:
+            return
+
+        nested: dict[str, Any] = event.payload
+        for key in reversed(list(event.entity.path.segments)):
+            nested = {key: nested}
+
+        if event.op == "update":
+            deep_merge_dict(self.data, nested)
+        elif event.op == "overwrite":
+            delete_nested_key(self.data, event.entity.path)
+            deep_merge_dict(self.data, nested)
+
+    def get_value(self, path: IpsoPath) -> VALUE_TYPES | None:
+        obj = self.data
+        for key in path.segments:
+            if not isinstance(obj, dict):
+                return None
+            obj = obj.get(key)
+        if isinstance(obj, dict):
+            return ValueField(**obj).value
+        return obj
+
+    def get_field(self, path: IpsoPath) -> ValueField | None:
+        obj = self.data
+
+        for key in path.segments:
+            if not isinstance(obj, dict):
+                return None
+            obj = obj.get(key)
+        return ValueField(**obj) if isinstance(obj, dict) else None
+
+    def build_write_value_obj(
+        self, path: IpsoPath, value: VALUE_TYPES
+    ) -> EgressMessageList:
+        payload = _value_to_payload(value)
+
+        return EgressMessageList(
+            [
+                Request(
+                    op="write",
+                    entity=Entity(device=self.id, path=path, service=self.service),
+                    payload=payload,
+                )
+            ]
+        )
+
+    def build_read_value_obj(self, path: IpsoPath) -> EgressMessageList:
+        return EgressMessageList(
+            [
+                Request(
+                    op="read",
+                    entity=Entity(device=self.id, path=path, service=self.service),
+                )
+            ]
+        )
+
+    @cached_property
+    def objects(self) -> dict[str, dict[str, IpsoObject]]:
+        result: dict[str, dict[str, IpsoObject]] = {}
+        for object_name, object_data in self.model_definition.objects.items():
+            multi_instance = object_data.get("multi_instance", False)
+            result[object_name] = {}
+
+            if multi_instance:
+                if object_name in self.data:
+                    for object_instance_id in self.data[object_name].keys():
+                        result[object_name][object_instance_id] = IpsoObject(
+                            object_name, object_instance_id, object_data
+                        )
+            else:
+                result[object_name]["0"] = IpsoObject(object_name, "0", object_data)
+        return result
+
+    def get_object(self, object_name: str, instance_id: str = "0") -> IpsoObject | None:
+        """Get a object by name and instance ID.
+
+        Args:
+            object_name: Name of the object (e.g., "actuator", "device")
+            instance_id: Instance ID (default: "0")
+
+        Returns:
+            DynamicObject if found, None otherwise
+        """
+        return self.objects.get(object_name, {}).get(instance_id)
+
+    def get_resource(
+        self, object_name: str, instance_id: str | int, resource_name: str
+    ) -> Any | None:
+        """Get a resource value from the device.
+
+        Args:
+            object_name: Name of the object (e.g., "actuator")
+            instance_id: Instance ID (e.g., "0" or 0)
+            resource_name: Name of the resource (e.g., "state")
+
+        Returns:
+            The resource value, or None if not found
+        """
+        obj = self.get_object(object_name, str(instance_id))
+        return obj.get_value(resource_name, self.data) if obj else None
+
+    def build_set_resource_obj(
+        self,
+        object_name: str,
+        instance_id: str | int,
+        resource_name: str,
+        value: Any,
+    ) -> EgressMessageList:
+        obj = self.get_object(object_name, str(instance_id))
+        if not obj:
+            raise ValueError(f"Object {object_name}[{instance_id}] not found")
+
+        resource = obj.get_resource(resource_name)
+        if not resource:
+            raise ValueError(
+                f"Resource {resource_name} not found in {object_name}[{instance_id}]"
+            )
+
+        if not resource.is_writable:
+            raise ValueError(
+                f"Resource {object_name}/{instance_id}/{resource_name} is not writable"
+            )
+
+        payload = _value_to_payload(value)
+        path = IpsoPath(
+            object_name=object_name,
+            object_instance_id=str(instance_id),
+            resource_name=resource_name,
+        )
+
+        return EgressMessageList(
+            [
+                Request(
+                    op="write",
+                    entity=Entity(device=self.id, path=path, service=self.service),
+                    payload=payload,
+                )
+            ]
+        )
+
+    def list_instances(self, object_name: str) -> list[str]:
+        """Get list of all instance IDs for an object.
+
+        Args:
+            object_name: Name of the object
+
+        Returns:
+            List of instance ID strings
+        """
+        return list(self.objects.get(object_name, {}).keys())
+
+    def __repr__(self) -> str:
+        return (
+            f"{self.__class__.__name__}(id={self.id!r}, "
+            f"model={self.model_definition.name!r}, "
+            f"type={self.model_definition.type!r})"
+        )
+
+
+class DeviceMap(RootModel[dict[str, Device]], MutableMapping):
+    def __add__(self, other: "DeviceMap") -> "DeviceMap":
+        return DeviceMap(self.root | other.root)
+
+    def __str__(self) -> str:
+        return self.model_dump_json(exclude_none=True)
+
+    def __getitem__(self, key: str) -> Device:
+        return self.root[key]
+
+    def __setitem__(self, key: str, value: Device) -> None:
+        self.root[key] = value
+
+    def __delitem__(self, key: str) -> None:
+        del self.root[key]
+
+    def __iter__(self):
+        return iter(self.root.keys())
+
+    def __len__(self) -> int:
+        return len(self.root)
+
+
+def build_discover_gen1_obj() -> EgressMessageList:
+    return EgressMessageList(
+        [
+            Request(
+                op="read",
+                entity=Entity(
+                    service="lemonbeatd", path=IpsoPath(object_name="devices")
+                ),
+            )
+        ]
+    )
+
+
+def build_discovery_obj() -> EgressMessageList:
+    return build_discover_gen1_obj()
--- a/gardena_smart_local_api/devices/device_builder.py
+++ b/gardena_smart_local_api/devices/device_builder.py
@@ -2,6 +2,11 @@ from typing import Any
 
 from ..messages import IngressMessageList, Reply
 from .device import Device, DeviceMap
+from .gen1 import WaterControl
+
+MODEL_NUMBER_MAP: dict[str, type[Device]] = {
+    "18869": WaterControl,
+}
 
 
 async def _create_device_from_data(device_data: dict[str, Any]) -> Device | None:
@@ -9,7 +14,7 @@ async def _create_device_from_data(device_data: dict[str, Any]) -> Device | None
     if model_number is None:
         return None
 
-    return await Device._from_raw(device_data)
+    return await MODEL_NUMBER_MAP.get(model_number, Device)._from_raw(device_data)
 
 
 async def create_devices_from_messages(messages: IngressMessageList) -> DeviceMap:
--- a/gardena_smart_local_api/devices/device_builder.py
+++ b/gardena_smart_local_api/devices/device_builder.py
@@ -1,0 +1,29 @@
+from typing import Any
+
+from ..messages import IngressMessageList, Reply
+from .device import Device, DeviceMap
+
+
+async def _create_device_from_data(device_data: dict[str, Any]) -> Device | None:
+    model_number = list(device_data.values())[0]["device"]["0"]["model_number"]["vs"]
+    if model_number is None:
+        return None
+
+    return await Device._from_raw(device_data)
+
+
+async def create_devices_from_messages(messages: IngressMessageList) -> DeviceMap:
+    devices: dict[str, Device] = {}
+    for msg in messages:
+        if not isinstance(msg, Reply):
+            continue
+        for device_id, device_data in msg.payload.items():
+            device = await _create_device_from_data({device_id: device_data})
+            if device is not None:
+                devices[device_id] = device
+    return DeviceMap(devices)
+
+
+async def create_devices_from_json(json_obj: str) -> DeviceMap:
+    msgs = IngressMessageList.model_validate_json(json_obj)
+    return await create_devices_from_messages(msgs)
--- a/gardena_smart_local_api/devices/gen1.py
+++ b/gardena_smart_local_api/devices/gen1.py
@@ -1,0 +1,296 @@
+from typing import ClassVar
+
+from pydantic import Field
+
+from ..messages import EgressMessageList, Entity, Request
+from ..model_loader import Gen1ModelDefinition
+from ..resources import IpsoPath
+from .device import Device
+
+
+class Gen1Device(Device):
+    model_definition: Gen1ModelDefinition = Field()
+    service: ClassVar[str] = "lemonbeatd"
+
+    def get_command(self, name: str) -> int:
+        return self.model_definition.commands[name]
+
+    def build_command_obj(self, command: int) -> EgressMessageList:
+        """Build a Gen1 command JSON structure.
+
+        Args:
+            command: Command code as int.
+
+        Returns:
+            EgressMessageList ready to be sent to the local GARDENA smart API.
+
+        Example:
+            >>> build_command_obj(3)
+            EgressMessageList([
+                Request(
+                    op="write",
+                    entity=Entity(
+                        device="device_id",
+                        path="lemonbeat/0/command",
+                        service="lemonbeatd"
+                    ),
+                    payload={"vi": 3}
+                )
+            ])
+        """
+        request = Request(
+            op="write",
+            entity=Entity(
+                device=self.id,
+                path=IpsoPath(
+                    object_name="lemonbeat",
+                    object_instance_id="0",
+                    resource_name="command",
+                ),
+                service=self.service,
+            ),
+            payload={"vi": command},
+        )
+        return EgressMessageList([request])
+
+    @property
+    def is_online(self) -> bool | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="connection_status",
+                object_instance_id="0",
+                resource_name="online",
+            )
+        )
+        if isinstance(value, bool):
+            return value
+        return None
+
+    def build_reload_online_status_obj(self) -> EgressMessageList:
+        return self.build_read_value_obj(
+            IpsoPath(
+                object_name="connection_status",
+                object_instance_id="0",
+                resource_name="online",
+            )
+        )
+
+    @property
+    def serial_number(self) -> str | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="device",
+                object_instance_id="0",
+                resource_name="serial_number",
+            )
+        )
+        return str(value) if value else None
+
+    @property
+    def manufacturer(self) -> str | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="device",
+                object_instance_id="0",
+                resource_name="manufacturer",
+            )
+        )
+        return str(value) if value else None
+
+    @property
+    def rf_link_quality(self) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="rf_link_quality",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
+
+    def build_reload_rf_link_quality_obj(self) -> EgressMessageList:
+        return self.build_read_value_obj(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="rf_link_quality",
+            )
+        )
+
+    @property
+    def rf_link_state(self) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="rf_link_state",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
+
+    def build_reload_rf_link_state_obj(self) -> EgressMessageList:
+        return self.build_command_obj(self.get_command("measure_rf_link"))
+
+    @property
+    def software_version(self) -> str | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="device",
+                object_instance_id="0",
+                resource_name="software_version",
+            )
+        )
+        return str(value) if value else None
+
+    @property
+    def hardware_version(self) -> str | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="device",
+                object_instance_id="0",
+                resource_name="hardware_version",
+            )
+        )
+        return str(value) if value else None
+
+
+class Gen1BatteryPoweredGen1Device(Gen1Device):
+    @property
+    def battery_level(self) -> float | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="battery_level",
+            )
+        )
+        if isinstance(value, (int, float)):
+            return float(value)
+        return None
+
+    def build_reload_battery_level_obj(self) -> EgressMessageList:
+        return self.build_command_obj(self.get_command("measure_battery"))
+
+
+class WaterControl(Gen1BatteryPoweredGen1Device):
+    def build_set_watering_timer_obj(self, seconds: int) -> EgressMessageList:
+        """Set the watering timer.
+
+        Args:
+            seconds: Duration of watering in seconds.
+
+        Returns:
+            EgressMessageList ready to be sent to the local GARDENA smart API.
+
+        Example:
+            >>> WaterControl(...).build_set_watering_timer_obj(3600)
+            EgressMessageList([
+                Request(
+                    request_id="2a8166c5-d60f-4ddd-8735-29aa3661a128",
+                    op="write",
+                    entity=Entity(
+                        device="device_id",
+                        path=IpsoPath(
+                            object_name="lemonbeat",
+                            object_instance_id="0",
+                            resource_name="watering_timer_1"
+                        ),
+                        service="lemonbeatd"
+                    ),
+                    payload={"vi": 3600}
+                )
+            ])
+        """
+        return self.build_write_value_obj(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="watering_timer_1",
+            ),
+            seconds,
+        )
+
+    def build_stop_watering_obj(self) -> EgressMessageList:
+        return self.build_set_watering_timer_obj(0)
+
+    @property
+    def watering_timer(self) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="watering_timer_1",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
+
+    @property
+    def is_opened(self) -> bool | None:
+        value = self.watering_timer
+        if value is not None:
+            return value > 0
+        return None
+
+    @property
+    def button_config_time(self) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="button_config_time",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
+
+    @property
+    def ambient_temperature(self) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="ambient_temperature",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
+
+    @property
+    def has_frost_warning(self) -> bool | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="frost_warning",
+            )
+        )
+        if isinstance(value, int):
+            return bool(value)
+        return None
+
+    def build_read_button_config_time_obj(self) -> EgressMessageList:
+        return self.build_read_value_obj(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="button_config_time",
+            )
+        )
+
+    def build_set_button_config_time_obj(self, seconds: int) -> EgressMessageList:
+        return self.build_write_value_obj(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="button_config_time",
+            ),
+            seconds,
+        )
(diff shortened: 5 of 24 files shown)


smart_system_local/devices/dynamic/device_builder.py
```diff
@@ -0,0 +1,332 @@
...
+
+        model_loader = await get_model_loader()
+        model_definition = model_loader.get_model(model_number)
+        if not model_definition:
+            raise ValueError(f"Unknown model number: {model_number}")
+
+        return cls(id=device_id, raw=device_data, model_definition=model_definition)
+
+    @staticmethod
+    def discover() -> Request:
```
easybe: This function returns the request to list all the devices for a given device service (hard-coded `lemonbeatd`). This does not seem like something that should be tied to a device.

Personally, I am not a big fan of static methods. IMHO classes should be used when there is state. There is nothing wrong with free functions.

smart_system_local/devices/dynamic/device_builder.py
```diff
@@ -0,0 +1,332 @@
...
+    def serial_number(self) -> str | None:
+        return self.get_value("device", "0", "serial_number")
+
+    @property
+    def firmware_version(self) -> str | None:
+        return self.get_value("device", "0", "firmware_version")
+
+    @property
+    def software_version(self) -> str | None:
+        return self.get_value("lemonbeat", "0", "software_version")
```
easybe: This is all lemonbeat-specific. So, at the very least this needs to be reflected in the class name/path. I don't know if this can/should be done in a generic way that works for lemonbeat and lwm2m devices, or even across device types. Perhaps just have classes for the data and move the logic somewhere else?

smart_system_local/devices/messages.py
```diff
@@ -0,0 +1,70 @@
...
+class Event(BaseModel):
+    """Event received from Smart System Local API."""
+
+    entity: Entity | None = None
+    metadata: ResponseMetadata | None = None
+    op: str | None = None
+    payload: dict | None = None
+
+
+class Response(BaseModel):
```
easybe:
```suggestion
class Reply(BaseModel):
```

as in request-reply pattern.

smart_system_local/devices/messages.py
```diff
@@ -0,0 +1,70 @@
+from pydantic import BaseModel
+
+
+class Entity(BaseModel):
+    """Entity information in Smart System Local responses."""
```
easybe:
```suggestion
    """Entity information in smart system local API responses."""
```

as in "GARDENA smart system"

easybe (commented):
I will need some more time to figure out for myself how I would implement this.

smart_system_local/devices/dynamic/device_builder.py
```diff
@@ -0,0 +1,332 @@
...
+        raise TypeError(f"Unsupported value type: {value_type}")
+
+
+class DynamicDevice(BaseModel):
+    id: str
+    raw: dict[str, Any] = Field(repr=False)
+    model_definition: ModelDefinition = Field()
+
+    @classmethod
+    async def from_raw(cls, raw_message: dict[str, Any]) -> "DynamicDevice":
```
easybe: I understand this coverers building an initial state for a device. We also need a way to update the state when an event comes in.

smart_system_local/devices/dynamic/device_builder.py
```diff
@@ -0,0 +1,332 @@
...
+            return {"ai": value}
+        else:
+            return {"as": value}
+    else:
+        raise TypeError(f"Unsupported value type: {value_type}")
+
+
+class DynamicDevice(BaseModel):
+    id: str
+    raw: dict[str, Any] = Field(repr=False)
```
easybe:
```suggestion
    state: dict[str, Any] = Field(repr=False)
```
or maybe
```suggestion
    data: dict[str, Any] = Field(repr=False)
```
?

smart_system_local/devices/dynamic/device_builder.py
```diff
@@ -0,0 +1,332 @@
...
+        obj = self.raw
+
+        for key in path:
+            if not isinstance(obj, dict):
+                return None
+            obj = obj.get(key)
+        return ValueField(**obj) if isinstance(obj, dict) else None
+
+    @property
+    def is_online(self) -> bool:
```
easybe: Do we need this and the following abstractions at this level? How about moving this to a higher abstraction layer where we distinguish between lemonbeat and lwm2m? Perhaps even leave it out for this PR.

smart_system_local/devices/dynamic/device_builder.py
```diff
@@ -0,0 +1,332 @@
...
+    elif isinstance(value, list):
+        if value and isinstance(value[0], int):
+            return {"ai": value}
+        else:
+            return {"as": value}
+    else:
+        raise TypeError(f"Unsupported value type: {value_type}")
+
+
+class DynamicDevice(BaseModel):
```
easybe:
```suggestion
class Device(BaseModel):
```

Why distinguish between dynamic and static? Can we move all of this one directory up?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  lint-and-test:
```
easybe:
```suggestion
  qa:
```
?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+    runs-on: ubuntu-latest
```
easybe:
```suggestion
    runs-on: ubuntu-22.04
```

smart_system_local/schema/device_models_schema.json
```diff
@@ -0,0 +1,174 @@
...
+          "description": "Whether multiple instances of this resource can exist"
+        },
+        "description": {
+          "type": "string",
+          "description": "Human-readable description"
+        }
+      }
+    }
+  }
+}
```
easybe: Missing final LF

gardena_smart_local/tests/data/devices.json
```diff
@@ -0,0 +1,188 @@
...
+    "metadata": { "sequence": 3500, "source": "lemonbeatd" },
+    "payload": {
+      "3034F8EE901267400000B333": {
+        "connection_status": {
+          "0": { "online": { "ts": 1768826098, "vb": true } },
+          "_urn": "urn:oma:lwm2m:x:28171"
+        },
+        "device": {
+          "0": {
+            "device_type": { "ts": 1769328781, "vs": "Sensor" },
```
easybe: Perhaps exclude non-WC data for now?

smart_system_local/devices/messages.py
```diff
@@ -0,0 +1,70 @@
...
+
+    path: str = ""
+    service: str | None = None
+    device: str | None = None
+
+
+class ResponseMetadata(BaseModel):
```
easybe: Perhaps remove the currently unused entries for now?

pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[project]
+name = "py-gardena-smart-system-local"
+version = "0.1.0"
+description = "Library to interact with the local Gardena api."
```
easybe:
```suggestion
description = "Library to interact with the local GARDENA API."
```

easybe (changes requested):

smart_system_local/devices/dynamic/device_builder.py
kelvan: this is only the low-level interface loaded from the yaml (you are right that the properties shouldn't be in there), the originally idea was to have a more high-level Devices class that implements the common properties.

Maybe something like WaterControl -> LemonbeatDevice -> Device (generated using the yaml schema) would be a good way?
On the other side do we want the ic24 and ic24 eco to be separate high-level classes or hide these low-level details there?

smart_system_local/devices/dynamic/device_builder.py
kelvan: this probably should be be added to a higher LemonbeatDevice level

.github/workflows/ci.yml
kelvan: Why not 24.04 LTS?

smart_system_local/devices/dynamic/device_builder.py
kelvan: moved to function `discover_lemonbeat_devices` and added `discover_devices` that returns a list of requests (currently only for lemonbeatd)

kelvan:
Is 120 ok as max-line-length?

gardena_smart_local/tests/data/devices.json
kelvan: depends if we need to add some "unknown device" handling in this PR, than it makes sense to have more devices in the test data.

.github/workflows/ci.yml
easybe: That is what I meant

easybe:
> Is 120 ok as max-line-length?

As I am a boomer :wink:, I like PEP 8's recommendation of 79. So, I would rather go with Ruff's default of 88. But, 120 is acceptable for me if that is what is preferred.

smart_system_local/devices/device_builder.py
```diff
@@ -0,0 +1,317 @@
...
+        """
+        return list(self.objects.get(object_name, {}).keys())
+
+    def __repr__(self) -> str:
+        return (
+            f"DynamicDevice(id={self.id!r}, model={self.model_definition.name!r}, type={self.model_definition.type!r})"
+        )
+
+
+def discover_lemonbeat_devices() -> Request:
```
easybe: In general, we should avoid using the terms "lemonbeat" and "lwm2m" in the code as they are technically not correct and could be confusing. Instead, we should use the terms "gen1" and "gen2".

easybe (changes requested):

easybe:
Instead of putting commits on top, please, amend previous commits. `git commit --fixup` and `git rebase -i` are your friends.

kelvan:
> > Is 120 ok as max-line-length?
> 
> As I am a boomer 😉, I like PEP 8's recommendation of 79. So, I would rather go with Ruff's default of 88. But, 120 is acceptable for me if that is what is preferred.

I changed it to the default 88, the docstrings were the only part too long

smart_system_local/devices/device_builder.py
kelvan: agree, I renamed the occurrences of lemonbeat/lwm2m to gen1/gen2

kelvan:
> Instead of putting commits on top, please, amend previous commits. `git commit --fixup` and `git rebase -i` are your friends.

I assumed we do squash merges of the PR, as the HA project does

gardena_smart_local_api/model_loader.py
```diff
@@ -0,0 +1,102 @@
...
+        if protocol in PROTOCOL_MAP:
+            return PROTOCOL_MAP[protocol](model_number=model_number, **data)
+        else:
+            raise ValueError(f"Unknown protocol: {protocol}")
+
+    async def load(self) -> None:
+        if self._loaded:
+            return
+
+        async with async_open(self.yaml_path, mode="r") as f:
```
easybe, line 60: Does this need to be async? We are just reading a file. I would opt for sync and make things simpler.

smart_system_local/schema/device_models_schema.json
```diff
@@ -0,0 +1,174 @@
+{
```
easybe: I don't see any reference to the schema file. How does this work? Or what would be the idea?

kelvan: the idea was to validate the yaml file, but we can remove it as we currently won't use it anyway

gardena_smart_local_api/model_loader.py
kelvan: as this will be part of the setup of the integration we should not make blocking calls (even if it is a short one)

smart_system_local/devices/dynamic/device_builder.py
kelvan: added a method to update the data, no deletion supported (I guess we will need it for the includable devices part)

easybe:
There are still some occurrences of "[Ll]emonbeat". The only valid usage is `lemonbeat` in `path` and `lemonbeatd` in `entity.service`.

gardena_smart_local/devices/device.py
```diff
@@ -100,7 +100,7 @@ def get_field(self, path: IpsoPath) -> ValueField | None:
             obj = obj.get(key)
         return ValueField(**obj) if isinstance(obj, dict) else None
 
-    def write_value(self, path: IpsoPath, value: int | str | bool | float) -> Request:
+    def write_value(self, path: IpsoPath, value: VALUE_TYPES) -> Request:
```
easybe: This change does not belong into this commit.

gardena_smart_local_api/messages.py
```diff
@@ -20,7 +20,7 @@ class Request(BaseModel):
     request_id: str | None = None
 
     def __str__(self) -> str:
-        return self.json(exclude_none=True)
+        return self.model_dump_json(exclude_none=True)
```
easybe, line 24: This seems also unrelated.

smart_system_local/devices/gen1.py
```diff
@@ -0,0 +1,216 @@
...
+                object_instance_id="0",
+                resource_name="online",
+            )
+        )
+
+    @property
+    def serial_number(self) -> str | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="evic",
```
easybe:
```suggestion
                object_name="device",
```

Something a test should catch?

gardena_smart_local/devices/device.py
```diff
@@ -0,0 +1,305 @@
...
+        payload = _value_to_payload(value)
+
+        return Request(
+            request_id=str(uuid.uuid4()),
+            op="write",
+            entity=Entity(device=self.id, path=path),
+            payload=payload,
+        )
+
+    def read_value(self, path: IpsoPath) -> Request:
```
easybe: As a user, the difference between `get_value` and `read_value` is not obvious from the name. IMHO it should be clear that a request is prepared/built/returned.

smart_system_local/devices/device.py
```diff
@@ -0,0 +1,305 @@
...
+def discover_gen1_devices() -> Request:
+    """Build a request to discover Gen1 devices."""
+    return Request(
+        request_id=str(uuid.uuid4()),
+        op="read",
+        entity=Entity(service="lemonbeatd", path=IpsoPath(object_name="devices")),
+    )
+
+
+def discover_devices() -> list[Request]:
```
easybe: The function does not discover devices. It merely prepares/creates/returns a request to do so.

easybe (changes requested):
Not a full review, but, some things I've noticed.

tests/test_gen1.py
```diff
@@ -0,0 +1,169 @@
...
+    """Test that Gen1 devices are created as Gen1Device instances."""
+    try:
+        from smart_system_local.devices.device_builder import create_device_from_data
+
+        device = cast(Gen1Device, await create_device_from_data(device_data))
+    except ValueError as e:
+        if "Unknown model number" in str(e) or "Could not extract model_number" in str(
+            e
+        ):
+            pytest.skip(f"Model not in schema: {e}")
```
easybe: Why do we want this?

gardena_smart_local/devices/gen1.py
```diff
@@ -0,0 +1,216 @@
...
+        )
+        if isinstance(value, (int, float)):
+            return int(value)
+        return None
+
+    def fetch_rf_link_state(self) -> Request:
+        return self.build_command(7)
+
+
+class Gen1BatteryPoweredMixIn(Gen1Device):
```
easybe: Do we really want to use mixins? IMHO they are considered bad practice nowadays -> Composition over inheritance. Still, I see why one would want to use them here. But, wouldn't we need use mixins in other places too and then things might get out of hand? Not sure if protocols could be helpful here.

gardena_smart_local/devices/device.py
```diff
@@ -0,0 +1,305 @@
...
+
+        if len(payload) != 1:
+            raise ValueError(
+                f"Expected exactly one device in payload, got {len(payload)}"
+            )
+
+        device_id, device_data = next(iter(payload.items()))
+
+        model_number = _extract_model_number(device_data)
+        if not model_number:
```
easybe:
```suggestion
        if model_number is None:
```

gardena_smart_local/devices/device.py
```diff
@@ -0,0 +1,305 @@
...
+            raise ValueError("Could not extract model_number from device data")
+
+        model_loader = await get_model_loader()
+        model_definition = model_loader.get_model(model_number)
+        if not model_definition:
```
easybe:
```suggestion
        if model_definition is None:
```

smart_system_local/model_loader.py
```diff
@@ -0,0 +1,102 @@
...
+        self._loaded = True
+
+    @property
+    def types(self) -> dict[str, str]:
+        return self._types
+
+    def get_model(self, model_number: str | int) -> ModelDefinition | None:
+        return self._models.get(str(model_number))
+
+    def get_model_by_type(self, device_type: str) -> list[ModelDefinition]:
```
easybe: Why is this needed?

smart_system_local/model_loader.py
```diff
@@ -0,0 +1,102 @@
...
+            f"name={self.name!r}, type={self.type!r})"
+        )
+
+
+class Gen1ModelDefinition(BaseModelDefinition):
+    protocol: Literal["gen1"] = "gen1"
+    commands: dict[str, int] = Field(default_factory=dict)
+
+
+class Gen2ModelDefinition(BaseModelDefinition):
```
easybe: I would add this when it is needed.

smart_system_local/model_loader.py
```diff
@@ -0,0 +1,102 @@
...
+    def types(self) -> dict[str, str]:
+        return self._types
+
+    def get_model(self, model_number: str | int) -> ModelDefinition | None:
+        return self._models.get(str(model_number))
+
+    def get_model_by_type(self, device_type: str) -> list[ModelDefinition]:
+        return [m for m in self._models.values() if m.type == device_type]
+
+    def list_models(self) -> list[ModelDefinition]:
```
easybe: Do we know that this will be needed?

gardena_smart_local/devices/gen1.py
```diff
@@ -0,0 +1,216 @@
...
+    @property
+    def rf_link_state(self) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="rf_link_state",
+            )
+        )
+        if isinstance(value, (int, float)):
```
easybe: Why `int` or `float`?

gardena_smart_local_api/devices/gen1.py
```diff
@@ -0,0 +1,216 @@
...
+                resource_name="watering_timer_1",
+            ),
+            seconds,
+        )
+
+    def stop_watering(self) -> Request:
+        return self.set_watering_timer(0)
+
+    @property
+    def button_config_time(self) -> int | None:
```
easybe, line 241: Do we need this (now)?

smart_system_local/devices/dynamic/device_builder.py
easybe: Thanks. IMHO this is something that should have a test.

gardena_smart_local/devices/device.py
```diff
@@ -0,0 +1,305 @@
...
+                    path=IpsoPath(
+                        object_name="lemonbeat",
+                        object_instance_id="0",
+                        resource_name="power_timer"
+                    )
+                )
+            )
+        """
+        return Request(
+            request_id=str(uuid.uuid4()),
```
easybe: Move into `Request`

gardena_smart_local/devices/gen1.py
```diff
@@ -0,0 +1,216 @@
...
+        return self.write_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="watering_timer_1",
+            ),
+            seconds,
+        )
+
+    def stop_watering(self) -> Request:
```
easybe:
```suggestion
    def build_stop_watering_request(self) -> Request:
```

gardena_smart_local/devices/gen1.py
```diff
@@ -0,0 +1,216 @@
...
+        request = super().read_value(path)
+        request.entity.service = "lemonbeatd"
+        return request
+
+    def write_value(self, path: IpsoPath, value: VALUE_TYPES) -> Request:
+        request = super().write_value(path, value)
+        request.entity.service = "lemonbeatd"
+        return request
+
+    def build_command(self, command: int) -> Request:
```
easybe:
```suggestion
    def build_command_request(self, command: int) -> Request:
```

pyproject.toml
```diff
@@ -0,0 +1,31 @@
+[project]
+name = "py-gardena-smart-system-local"
```
easybe:
```suggestion
name = "gardena-smart-local-api"
```

pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+version = "0.1.0"
+description = "Library to interact with the local GARDENA API."
```
easybe:
```suggestion
description = "Library to interact with the local GARDENA smart API."
```

README.md
```diff
@@ -0,0 +1,22 @@
+# py-gardena-smart-system-local
```
```suggestion
# Python library for GARDENA smart local API

Enables controlling and monitoring GARDENA smart devices in the local network, without going through the cloud. 
```

gardena_smart_local/messages.py
```diff
@@ -0,0 +1,50 @@
+from pydantic import BaseModel
```
easybe: Move non-device one level up.

gardena_smart_local_api/devices/device.py
```diff
@@ -0,0 +1,305 @@
...
+
+from ..model_loader import ModelDefinition, get_model_loader
+from .messages import Entity, Request
+from .resources import IpsoObject, IpsoPath, ValueField
+
+VALUE_TYPES = bool | int | float | str | list[int] | list[str]
+
+
+def _value_to_payload[T: VALUE_TYPES](value: T) -> dict[str, T]:
+    """Convert a Python value to the appropriate payload format.
```
easybe: Remove docstrings for internal functions (unless some explanation is needed) and add docstrings for public API.

tests/test_gen1.py
```diff
@@ -0,0 +1,153 @@
+import json
```
easybe: I would move the tests one level up.

gardena_smart_local_api/devices/device.py
```diff
@@ -0,0 +1,224 @@
...
+        return None
+
+
+class Device(BaseModel):
+    id: str
+    data: dict[str, Any] = Field(repr=False)
+    model_definition: ModelDefinition = Field()
+
+    @classmethod
+    async def from_raw(cls, raw_message: dict[str, Any]) -> "Device":
```
easybe: This function should only take, e.g.:
```json
"3034F8EE90126D400001737D": {
  "connection_status": {
    "0": { "online": { "ts": 1761915305, "vb": true } },
    "_urn": "urn:oma:lwm2m:x:28171"
  },
  "device": {
    "0": {
      "device_type": { "ts": 1769328850, "vs": "Water Control" },
      "error_code": { "ai": [0], "ts": 1769328850 },
      "firmware_version": { "ts": 1769328850, "vs": "4.0.0-1.6.1" },
      "hardware_version": { "ts": 1769328850, "vs": "0.3.5" },
      "manufacturer": { "ts": 1769328850, "vs": "Gardena" },
      "model_number": { "ts": 1769328850, "vs": "18869" },
      "serial_number": { "ts": 1769328850, "vs": "00095101" },
      "software_version": { "ts": 1769328850, "vs": "2.5.0" },
      "supported_binding_and_modes": { "ts": 1769328850, "vs": "U" },
      "utc_offset": { "ts": 1768845002, "vs": "UTC+01:00" }
    },
    "_urn": "urn:oma:lwm2m:oma:3:1.1"
  },
(...)
```

Supplying a response from the `devices` command should not be supported IMHO.

gardena_smart_local_api/devices/device_builder.py
```diff
@@ -0,0 +1,45 @@
...
+        try:
+            if device_data:
+                first_value = next(iter(device_data.values()))
+                return first_value["device"]["0"]["model_number"]["vs"]
+        except (KeyError, TypeError, StopIteration):
+            pass
+        return None
+
+
+async def create_device_from_data(device_data: dict[str, Any]) -> Device:
```
easybe: This function should take the response of the `devices` command and return a list of devices.

```suggestion
async def create_devices_from_reply(reply: Reply) -> List[Device]:
```

gardena_smart_local_api/devices/device_builder.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def extract_device_id(message: Event | Reply) -> str | None:
+    # Try to get device ID from entity first
+    if message.entity and message.entity.device:
+        return message.entity.device
+
+    # Try to extract from payload (single device)
+    if hasattr(message, "payload") and message.payload:
+        if len(message.payload) == 1:
```
easybe: We should not support this, it  not very intuitive.

gardena_smart_local_api/devices/device_builder.py
```diff
@@ -0,0 +1,45 @@
...
+
+async def create_device_from_data(device_data: dict[str, Any]) -> Device:
+    model_number = _extract_model_number(device_data)
+    if not model_number:
+        raise ValueError("Could not extract model_number from device data")
+
+    return await MODEL_NUMBER_MAP.get(model_number, Device).from_raw(device_data)
+
+
+def extract_device_id(message: Event | Reply) -> str | None:
```
easybe: Is this needed to e.g. associate an `Event` with a `Device` object for update?

easybe:
More feedback can be found in #7

gardena_smart_local_api/devices/device.py
easybe: Actually, I don't think a dict with one element is a good idea. Pass the ID as separate argument?

gardena_smart_local_api/devices/gen1.py
easybe: @kelvan would like to keep it.

gardena_smart_local_api/devices/device_builder.py
```diff
@@ -0,0 +1,63 @@
...
+
+async def create_device_from_data(device_data: dict[str, Any]) -> Device:
+    model_number = _extract_model_number(device_data)
+    if not model_number:
+        raise ValueError("Could not extract model_number from device data")
+
+    return await MODEL_NUMBER_MAP.get(model_number, Device).from_raw(device_data)
+
+
+async def create_devices_from_event(event: Event) -> list[Device]:
```
easybe: Why from event? Do we already support inclusion events? Otherwise, we should only create devices from replies of the `devices` request, no?

easybe (commented):
And some previous comments still need to be addressed/answered.

gardena_smart_local_api/messages.py
```diff
@@ -0,0 +1,52 @@
...
+
+class Entity(BaseModel):
+    """Entity information in smart system local API responses."""
+
+    path: IpsoPath
+    service: str | None = None
+    device: str | None = None
+
+
+class Request(BaseModel):
```
easybe, line 17: Can we introduce a container that serializes to a JSON array? E.g.:
* MessageList
* MessageContainer
* Frame
* Messages
* Envelope
* ...

See https://github.com/cloudless-garden/gardena-smart-local/pull/7#discussion_r2842139701

easybe:
```python
class Message(BaseModel):
    pass

class Request(Message):
    (...)

class MessageList(BaseModel):
    messages = List[Message]
```

gardena_smart_local_api/devices/device.py
```diff
@@ -0,0 +1,228 @@
...
+        payload = raw_message.get("payload") or raw_message
+
+        if len(payload) != 1:
+            raise ValueError(
+                f"Expected exactly one device in payload, got {len(payload)}"
+            )
```
LukasWoodtli: This method should receive only the data of one device. Extracting the payload and retrieving the devices data should be done outside that method.

gardena_smart_local_api/devices/device.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def __repr__(self) -> str:
+        return (
+            f"{self.__class__.__name__}(id={self.id!r}, "
+            f"model={self.model_definition.name!r}, "
+            f"type={self.model_definition.type!r})"
+        )
+
+
+def build_discover_gen1_request() -> Request:
```
LukasWoodtli: If this is public API it should return a list of `Request`. This is expected to be sent over WS. Alternatively use just build_discover_requests() as public API and make this function private (the better option IMHO, as the user should not be concerned about device generations).

gardena_smart_local_api/devices/device.py
```diff
@@ -0,0 +1,228 @@
...
+    """Build a request to discover Gen1 devices."""
+    return Request(
+        op="read",
+        entity=Entity(service="lemonbeatd", path=IpsoPath(object_name="devices")),
+    )
+
+
+def build_discover_requests() -> list[Request]:
+    """Build requests to discover devices for all supported protocols."""
+    return [build_discover_gen1_request()]
```
LukasWoodtli: I think this is not easily JSON serializable. As it is a list of Models. Maybe also add a helper function for creating the JSON (as it's needed for sending over WS).

gardena_smart_local_api/devices/device_builder.py
```diff
@@ -0,0 +1,63 @@
...
+
+MODEL_NUMBER_MAP: dict[str, type[Device]] = {
+    "18869": WaterControl,
+}
+
+
+def _extract_model_number(device_data: dict[str, Any]) -> str | None:
+    try:
+        return device_data["device"]["0"]["model_number"]["vs"]
+    except (KeyError, TypeError):
```
LukasWoodtli: Is this really needed? The JSON always has the same format.

gardena_smart_local_api/devices/gen1.py
```diff
@@ -0,0 +1,293 @@
...
+                object_instance_id="0",
+                resource_name="battery_level",
+            )
+        )
+        if isinstance(value, (int, float)):
+            return float(value)
+        return None
+
+    def build_reload_battery_level_request(self) -> Request:
+        # TODO: add command to yaml schema
```
LukasWoodtli: Why the TODO. Fix it and remove the comment.

gardena_smart_local_api/devices/gen1.py
```diff
@@ -0,0 +1,293 @@
...
+        """Build a Gen1 command JSON structure.
+
+        Args:
+            command: Command code as int.
+
+        Returns:
+            Request ready to be sent to the local GARDENA smart API.
+
+        Example:
+            >>> device.build_command_request(3)
```
LukasWoodtli: We should add `doctest` to `pytest` and CI.

```
# content of pytest.ini
[pytest]
addopts = --doctest-modules
```

See: https://docs.pytest.org/en/7.1.x/how-to/doctest.html

gardena_smart_local_api/devices/device_builder.py
easybe: I guess we should be able to get or update a device based on event data.

LukasWoodtli (commented):
Looks already quite good!

LukasWoodtli:
> > Is 120 ok as max-line-length?
> 
> As I am a boomer 😉, I like PEP 8's recommendation of 79. So, I would rather go with Ruff's default of 88. But, 120 is acceptable for me if that is what is preferred.

To be honest, you are more like Gen Y.

I'm not so fundamentalistic about line length...

gardena_smart_local_api/devices/device.py
```diff
@@ -0,0 +1,228 @@
...
+from pydantic import BaseModel, Field
+
+from ..ipso import canonical_ipso_payload
+from ..messages import Entity, Request
+from ..model_loader import ModelDefinition, get_model_loader
+from ..resources import VALUE_TYPES, IpsoObject, IpsoPath, ValueField
+from ..utils import deep_merge_dict
+
+
+def _value_to_payload[T: VALUE_TYPES](value: T) -> dict[str, T]:
```
LukasWoodtli, line 13: We have the mapping specified in 'device_models.yaml'. Can't we reuse that information?

gardena_smart_local_api/model_loader.py
LukasWoodtli: We also should not use blocking I/O in an async function.

gardena_smart_local_api/messages.py
LukasWoodtli: MessageList will serialize to a JSON object with a field 'messages' that has a list of messages. So a custom de-/serializer would be required.

easybe:
> > > Is 120 ok as max-line-length?
> > 
> > 
> > As I am a boomer 😉, I like PEP 8's recommendation of 79. So, I would rather go with Ruff's default of 88. But, 120 is acceptable for me if that is what is preferred.
> 
> To be honest, you are more like Gen Y.

IMHO: boomer == old school

https://www.reddit.com/r/Millennials/comments/166clv3/the_term_boomer/

gardena_smart_local_api/devices/device_builder.py
```diff
@@ -0,0 +1,36 @@
...
+from .gen1 import WaterControl
+
+MODEL_NUMBER_MAP: dict[str, type[Device]] = {
+    "18869": WaterControl,
+}
+
+
+async def create_device_from_data(device_data: dict[str, Any]) -> Device:
+    model_number = list(device_data.values())[0]["device"]["0"]["model_number"]["vs"]
+    if not model_number:
```
easybe: You seem to be checking for an empty string or None. Does that make sense? Isn't it more likely to get an exception here?

gardena_smart_local_api/messages.py
```diff
@@ -0,0 +1,50 @@
...
+    payload: dict | None = None
+
+
+class Reply(BaseModel):
+    """Reply from a smart system local API command execution."""
+
+    request_id: str
+    success: bool | None = None
+    entity: Entity | None = None
+    payload: dict | None = None
```
easybe: I've update https://github.com/cloudless-garden/gardena-smart-local-api/pull/7. I like the API much better now. I think the following would be helpful too. Or, how would you solve 

https://github.com/cloudless-garden/gardena-smart-local-api/blob/b450a0658dca86c2f0c5bc23873ab41697818c85/examples/water-control.py#L51-L54
 
?

```suggestion
    payload: dict | None = None
    
    
class ReplyList(RootModel[list[Reply]]):
    def __add__(self, other: "ReplyList") -> "ReplyList":
        return ReplyList(self.root + other.root)

    def __str__(self) -> str:
        return self.model_dump_json(exclude_none=True)

    def __getitem__(self, key: int) -> Reply:
        return self.root[key]

    def __iter__(self):
        return iter(self.root)

    def __len__(self) -> int:
        return len(self.root)
```

And the same for `Event` I guess.

pyproject.toml
```diff
@@ -0,0 +1,39 @@
+[project]
+name = "gardena-smart-local-api"
+version = "0.1.0"
```
easybe:
```suggestion
version = "0.0.1"
```

gardena_smart_local_api/devices/device_builder.py
```diff
@@ -0,0 +1,36 @@
...
+        raise ValueError("Could not extract model_number from device data")
+
+    return await MODEL_NUMBER_MAP.get(model_number, Device).from_raw(device_data)
+
+
+async def create_devices_from_reply(event: Reply) -> DeviceList:
```
easybe: To cover this uses case

https://github.com/cloudless-garden/gardena-smart-local-api/blob/b450a0658dca86c2f0c5bc23873ab41697818c85/examples/water-control.py#L51-L54

I guess I would rather want

```suggestion
async def create_devices_from_json(json: str) -> DeviceList:
```

after all.

gardena_smart_local_api/messages.py
```diff
@@ -0,0 +1,102 @@
...
+def validate_message(data: dict) -> Message:
+    """
+    Validate incoming message data and return as Event or Reply.
+    Args:
+        data: The raw message data as a dictionary
+    Returns:
+        An instance of Event or Reply depending on the message type
+    Raises:
+        ValidationError: If the data does not conform to none of the Message schemas
+    """
```
easybe:
```suggestion
    """Validate incoming message data and return as Event or Reply.

    Args:
        data: The raw message data as a dictionary
    Returns:
        An instance of Event or Reply depending on the message type
    Raises:
        ValidationError: If the data does not conform to none of the Message schemas
    """
```

See: https://peps.python.org/pep-0257/

gardena_smart_local_api/devices/device_builder.py
```diff
@@ -0,0 +1,36 @@
...
+                device = await create_device_from_data({device_id: device_data})
+                devices[device_id] = device
+            except ValueError as e:
+                # Log the error and skip this device
+                print(f"Error creating device from data: {e}")
+    return DeviceList(devices)
+
+
+async def create_devices_from_data(data: str) -> DeviceList:
+    return await create_devices_from_reply(Reply.model_validate_json(data))
```
easybe: IMHO not good enough yet, as `data` (directly from the WebSocket) is a JSON array.

gardena_smart_local_api/devices/device.py
```diff
@@ -0,0 +1,258 @@
...
+        value: Value to convert (bool, int, float, str, or list)
+
+    Returns:
+        Dictionary with appropriate type key (vb, vi, vf, vs)
+
+    Raises:
+        TypeError: If value type is not supported
+    """
+    TYPE_MAP = {
+        bool: "vb",
```
broglep-work, line 15: we should consider adding all types the protocol supports. please also keep in mind that the current implementation will not work in all cases, as there is not a 1:1 mapping between python types and protocol types. at least we should add a notice / todo, that this is not fully complete implementation just yet

gardena_smart_local_api/devices/device.py
```diff
@@ -0,0 +1,258 @@
...
+            raise ValueError("Could not extract model_number from device data")
+
+        model_loader = await get_model_loader()
+        model_definition = model_loader.get_model(model_number)
+        if model_definition is None:
+            raise ValueError(f"Unknown model number: {model_number}")
+
+        return cls(id=device_id, data=device_data, model_definition=model_definition)
+
+    def update_data(self, new_data: dict[str, Any]) -> None:
```
broglep-work: this interface is not sufficient for correctly merging data (from an protocol message). We are missing the operation type, most notably the operation "overwrite" will not properly work with the current deep merge implementation and cause outdated data. Delete is the other case that would require special handling. 

Please also consider adding proper type annotation. the implementation of this method expects a protocol message, and not arbitrary data (update_data is very confusing and could lead consumer to think this is for updating device data using canonical payload format)

broglep-work (dismissed):

gardena_smart_local_api/devices/device_builder.py
kelvan: You could not add an array of messages into it anyway, even if it is currently always only an array with one item, it could theoretically contain different message types, so it would not be possible to return a DeviceList for this.

gardena_smart_local_api/devices/device_builder.py
kelvan: I would rather enforce using dataclasses as soon as possible to have the data structure validated. A json array from the websocket could contain mixed types, at least theoretically for future implementation of the websocketd. Not sure I want to work around the quirk that the array always only contain one element.
Otherwise I would need to assume all elements are the same type and implement a dataclass for a list representation or always only take the first element and miss messages if it contains multiple elements.

gardena_smart_local_api/messages.py
```diff
@@ -0,0 +1,90 @@
...
+    Validate incoming message data and return as ErrorMessage, Reply, or Event.
+
+    Args:
+        data: The raw message data as a dictionary
+    Returns:
+        An instance of ErrorMessage, Reply, or Event depending on the message type
+    Raises:
+        ValueError: If the data does not conform to any of the Message schemas
+    """
+    for message_cls in (ErrorMessage, Reply, Event):
```
easybe: Sort by likeliness:
```suggestion
    for message_cls in (Event, Reply, ErrorMessage):
```

gardena_smart_local_api/messages.py
easybe: Also, `ErrorMessage` is not mentioned.

gardena_smart_local_api/messages.py
```diff
@@ -0,0 +1,99 @@
...
+    entity: Entity
+    op: str
+    payload: dict | None = None
+    request_id: str | None = Field(default_factory=lambda: str(uuid.uuid4()))
+
+    def __str__(self) -> str:
+        return self.model_dump_json(exclude_none=True)
+
+
+class RequestList(RootModel[list[Request]]):
```
easybe:
```suggestion
class EgressMessageList(RootModel[list[Request]]):
```

gardena_smart_local_api/devices/device.py
```diff
@@ -0,0 +1,265 @@
...
+                op="read",
+                entity=Entity(
+                    service="lemonbeatd", path=IpsoPath(object_name="devices")
+                ),
+            )
+        ]
+    )
+
+
+def build_discovery_json() -> RequestList:
```
easybe: I find the name somewhat confusing now, as the following does not work:
```
json.dumps(build_discovery_json())
```

This would not be the case if we would return the output of `.model_dump()`.

Alternatively, we could just rename the function to e.g. `build_discovery_obj`.

As discussed, we should implement what is best for the integration. Convenience functions can always be added later.

Functions like `build_set_watering_timer_request()` should then be adapted accordingly.

README.md
```diff
@@ -0,0 +1,72 @@
...
+### Creating devices from JSON
+
+`create_devices_from_json` parses a JSON reply from the local GARDENA API and returns a `DeviceMap` (a dict-like container keyed by device ID).
+
+```python
+import asyncio
+from gardena_smart_local_api.devices import create_devices_from_json, WaterControl
+
+async def main():
+    with open("reply.json") as f:
```
easybe: Instead of loading form a file I would rather want to see a WebSocket example.

Perhaps we should rather merge https://github.com/cloudless-garden/gardena-smart-local-api/pull/7 and reference to that. Runable code is also easier to maintain.

README.md
```diff
@@ -0,0 +1,72 @@
...
+            print(f"  Valve open:  {device.is_opened}")
+
+asyncio.run(main())
+```
+
+### Controlling a Water Control device
+
+```python
+# Start watering for 30 minutes
+requests = device.build_set_watering_timer_request(seconds=1800)
```
easybe: I don't find this very usefully without more context. See other comment.

easybe (changes requested):

gardena_smart_local_api/devices/device.py
easybe: We'll go with `build_discovery_obj` and `build_set_watering_timer_obj`

easybe (approved):
Good enough (for a first version) I would say 👍

Thanks for bearing with me!

PR #6, opened 2026-02-12
Add basic gitignore

kelvan:


.gitignore
```diff
@@ -0,0 +1,61 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# Distribution / packaging
```
easybe, line 5: Will our build system create all of these directories?

.gitignore
```diff
@@ -0,0 +1,61 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
```
easybe: In my experience ignoring `*.pyc` is enough. No need to list the directory.

.gitignore
```diff
@@ -0,0 +1,61 @@
...
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# Sphinx documentation
+docs/_build/
```
easybe: We currently don't use Sphinx

.gitignore
```diff
@@ -0,0 +1,61 @@
...
+.idea/
+.vscode/
+*.swp
+*.swo
+*~
+.claude/
+
+# Linters
+.pytest_cache/
+.ruff_cache/
```
easybe: Missing LF. Best to configure your editor to add it on save.

easybe (changes requested):
If this is a standard template that we want to blindly use, I would like to see the source in the commit message body. Otherwise, I would prefer a minimal `.gitignore` tailored to our needs.

.gitignore
kelvan: on unix yes, not sure if this is true for Windows, but not sure if we should care about Windows.

.gitignore
kelvan: reduced it to the folders created by uv and pip install

.gitignore
```diff
@@ -0,0 +1,41 @@
...
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+.python-version
+uv.lock
+
+# OS X stuff
```
easybe:
```suggestion
# macOS stuff
```

.gitignore
```diff
@@ -0,0 +1,41 @@
...
+.DS_Store
+.AppleDouble
+.LSOverride
+Icon
```
easybe: I am a Mac user and have never had issues with these files. I don't really see the need to add these unless someone encounters them on their machine. Should they end up in a PR, I will spot them.

.gitignore
```diff
@@ -0,0 +1,41 @@
...
+
+# Logs
+*.log
+pip-log.txt
```
easybe: Why would we expect these files?

LukasWoodtli:
Squash this into one commit. There is no need to keep them seperate.

.gitignore
kelvan: I'm not a Mac user, but seen `.DS_Store` so often all over the place

easybe: Yes, `.DS_Store` files should be ignored, but not the other ones as they are not as common.

easybe (approved):

PR #7, opened 2026-02-23
Add valve example

easybe:
Run with:

    uv sync --group examples
    uv run examples/water-control.py --help

---

Depends on: https://github.com/cloudless-garden/gardena-smart-local-api/pull/15

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
 
 jobs:
   qa:
--- a/README.md
+++ b/README.md
@@ -14,6 +14,19 @@ Enables controlling and monitoring GARDENA smart devices in the local network, w
 pip install gardena-smart-local-api
 ```
 
+## Using the Library
+
+Have a look at our [example code][examples].
+
+You can run the examples from within the repository as follows:
+
+```txt
+uv sync --group examples
+uv run gardena_smart_local_api/examples/valve.py --help
+```
+
+[examples]: https://github.com/cloudless-garden/gardena-smart-local-api/tree/main/gardena_smart_local_api/examples
+
 ## Contributing
 
 ### Commits
--- a/gardena_smart_local_api/examples/__init__.py
+++ b/gardena_smart_local_api/examples/__init__.py
@@ -1,0 +1,3 @@
+from .app import ExampleApp
+
+__all__ = ["ExampleApp"]
--- a/gardena_smart_local_api/examples/app.py
+++ b/gardena_smart_local_api/examples/app.py
@@ -1,0 +1,121 @@
+import argparse
+import asyncio
+import base64
+import ssl
+from collections.abc import Iterable
+from typing import Self
+
+import websockets
+
+from gardena_smart_local_api.devices import (
+    DeviceMap,
+    build_discovery_obj,
+    create_devices_from_messages,
+)
+from gardena_smart_local_api.messages import (
+    EgressMessageList,
+    IngressMessageList,
+    Reply,
+)
+
+
+class ExampleApp:
+    def __init__(self, extra_args: Iterable[dict] = ()):
+        parser = argparse.ArgumentParser()
+        parser.add_argument(
+            "-g",
+            "--gateway",
+            required=True,
+            help="e.g. GARDENA-123456 or IP address",
+        )
+        parser.add_argument(
+            "-p",
+            "--password",
+            required=True,
+            help="First 8 characters of ID found on the back of the gateway",
+        )
+        parser.add_argument(
+            "-d",
+            "--device-id",
+            required=False,
+            help='ID of device to control (list IDs with "list" command)',
+        )
+
+        for arg_info in extra_args:
+            parser.add_argument(
+                *arg_info["name_or_flags"],
+                **{k: v for k, v in arg_info.items() if k != "name_or_flags"},
+            )
+
+        self.parser = parser
+        self.ws = None
+        self.devices = DeviceMap({})
+
+    async def connect(self):
+        self.args = self.parser.parse_args()
+        uri = f"wss://{self.args.gateway}:8443"
+        password = self.args.password
+        auth = base64.b64encode(f"_:{password}".encode()).decode("ascii")
+        headers = {"Authorization": f"Basic {auth}"}
+
+        ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        ssl_context.check_hostname = False
+        ssl_context.verify_mode = ssl.CERT_NONE
+
+        self.ws = await websockets.connect(
+            uri,
+            ssl=ssl_context,
+            additional_headers=headers,
+            close_timeout=0,
+        )
+
+    async def disconnect(self):
+        if self.ws:
+            await self.ws.close()
+            self.ws = None
+
+    async def send(self, data: str):
+        if not self.ws:
+            raise RuntimeError("Not connected")
+        await self.ws.send(data)
+
+    async def receive(self) -> str | None:
+        if not self.ws:
+            raise RuntimeError("Not connected")
+        try:
+            return await asyncio.wait_for(self.ws.recv(), timeout=3.0)
+        except TimeoutError:
+            pass
+        return None
+
+    async def send_request(self, msgs: EgressMessageList) -> IngressMessageList | None:
+        await self.send(str(msgs))
+
+        request_ids = {m.request_id for m in msgs}
+
+        for _ in range(10):
+            reply_json = await self.receive()
+            if reply_json is None:
+                continue
+            replies = IngressMessageList.model_validate_json(reply_json)
+            if not replies or not isinstance(replies[0], Reply):
+                continue
+            if {r.request_id for r in replies} == request_ids:
+                return replies
+
+        return None
+
+    async def discover_devices(self) -> DeviceMap:
+        discovery = build_discovery_obj()
+        replies = await self.send_request(discovery)
+        if replies is None:
+            return DeviceMap({})
+        return await create_devices_from_messages(replies)
+
+    async def __aenter__(self) -> Self:
+        await self.connect()
+        self.devices = await self.discover_devices()
+        return self
+
+    async def __aexit__(self, *args):
+        await self.disconnect()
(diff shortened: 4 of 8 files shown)


examples/water-control.py
```diff
@@ -0,0 +1,62 @@
...
+
+import argparse
+import asyncio
+import base64
+import json
+import ssl
+
+import websockets
+
+from gardena_smart_local.devices.device import build_discover_gen1_request
```
easybe: I would prefer this:

```suggestion
from gardena_smart_local.devices import build_discover_gen1_request
```

examples/water-control.py
```diff
@@ -0,0 +1,62 @@
...
+from gardena_smart_local.devices.device_builder import create_device_from_data
```
easybe:
```suggestion
from gardena_smart_local.devices import create_device_from_data
```

examples/water-control.py
```diff
@@ -0,0 +1,62 @@
...
+        msgs = json.loads(raw)
+        reply = Reply.model_validate_json(json.dumps(msgs[0]))
+
+        print(reply.model_dump())
+
+        for dev in reply.payload:
+            print(f"Device ID: {dev}")
+
+        # devices = create_devices_from_reply(reply)
+        # wc = devices[args.device_id]
```
easybe: I would prefer an API like this

examples/water-control.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+async def main():
+    async with websockets.connect(uri, ssl=ssl_context, extra_headers=headers) as ws:
+        request = build_discover_gen1_request()
+        msgs = json.dumps([request.model_dump(exclude_none=True)])
+        await ws.send(msgs)
+
+        raw = await asyncio.wait_for(ws.recv(), timeout=3.0)
+        msgs = json.loads(raw)
```
easybe: We need to find a way to abstract away the array of messages.

easybe: https://github.com/cloudless-garden/gardena-smart-local/pull/5#discussion_r2857717101

examples/water-control.py
LukasWoodtli: Agree!

examples/water-control.py
```diff
@@ -0,0 +1,74 @@
...
+        # devices = create_devices_from_reply(reply)
+        # wc = devices[args.device_id]
+        wc = await create_device_from_data(
+            args.device_id, reply.payload[args.device_id]
+        )
+
+        print(f"Watering with {wc.id}...")
+
+        request = wc.build_set_watering_timer_request(10)
+        msgs = json.dumps([request.model_dump(exclude_none=True)])
```
LukasWoodtli: Serialization to JSON should also encapsulated.

examples/water-control.py
kelvan: added DeviceList and RequestList models that behave like dicts, not sure if it is the best name, Devices/Requests could lead to easy misreads

examples/water-control.py
kelvan: added `class RequestList(RootModel[list[Request]])` similar to the RequestEnvelop, it also allows addition requests1+requests2 -> new RequestList

examples/water-control.py
kelvan: the methods now return RequestList objects that can be (de)serialized directly

examples/requirements.txt
```diff
@@ -0,0 +1,2 @@
+gardena-smart-local-api
```
easybe: Should we rather do this the `uv` way? How would that look like?

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
       - name: Run type checking
         run: |
-          uv run ty check
+          uv run ty check gardena_smart_local_api
```
kelvan: why excluding the tests and examples from type checks?
also if we want to exclude folders pyproject.toml would be the better place to define it imho

examples/water-control.py
```diff
@@ -0,0 +1,83 @@
...
+    "-p",
+    "--password",
+    required=True,
+    help="First 8 characters of ID found on the back of the gateway",
+)
+parser.add_argument(
+    "-d",
+    "--duration",
+    required=False,
+    default=10,
```
kelvan: maybe use `type=int` to validate the arg is an integer

examples/water-control.py
```diff
@@ -0,0 +1,83 @@
...
+
+        if not args.device_id:
+            print("No device_id provided. Exiting...")
+            return
+
+        wc = devices[args.device_id]
+
+        print(f"Watering with {wc.id}...")
+
+        request = wc.build_set_watering_timer_obj(int(args.duration))
```
kelvan: int cast not needed if type defined in argument

README.md
```diff
@@ -10,22 +10,31 @@ Enables controlling and monitoring GARDENA smart devices in the local network, w
...
+```
+
+### Linting
+
+```txt
 uv sync --all-groups
 uv run ruff check
 uv run ruff format --check
-uv run ty check
+uv run ty check gardena_smart_local_api
```
kelvan: as above, why excluding tests and examples from type checks?

README.md
```diff
@@ -10,22 +10,31 @@ Enables controlling and monitoring GARDENA smart devices in the local network, w
...
 ```
 
-## Running Tests
+### Running Tests
 
-```bash
+```txt
```
kelvan, line 64: why use a text block for shell code?

examples/water-control.py
```diff
@@ -0,0 +1,83 @@
...
+
+async def main():
+    async with websockets.connect(
+        uri,
+        ssl=ssl_context,
+        additional_headers=headers,
+    ) as ws:
+        discovery = build_discovery_obj()
+        await ws.send(str(discovery))
+        reply_json = await asyncio.wait_for(ws.recv(), timeout=3.0)
```
kelvan: should we validate if the reply has the correct request_id set? if an update message got pushed at this moment it would fail

examples/water-control.py
```diff
@@ -0,0 +1,83 @@
...
+import ssl
+
+import websockets
+
+from gardena_smart_local_api.devices import (
+    build_discovery_obj,
+    create_devices_from_json,
+)
+
+parser = argparse.ArgumentParser()
```
kelvan: missing a description what the command does, the script name suggests controlling a water control, but it only covers opening the valve for a specific time, not for example closing a valve.

examples/water-control.py
kelvan, line 1: do we want dashes in filenames, as the file is not imported it is not a real problem, but underscores are more common

README.md
easybe: See reasoning in commit message.

Here an example of highlighting that does not make sense:

image

Here an example where it makes sense:
```bash
if test $# -eq 2; then
    echo "Hello"
fi
```

README.md
easybe: I did not find it necessary for the examples, but I guess I can fix the code instead.

examples/water-control.py
easybe: I have had the same argument in the past and therefore started using `-` more often with GARDENA, but, I don't really care, I'll change it.

README.md
kelvan: I just see a risk that the examples could break without anyone notice, basic type check would prevent some of these problems

examples/water-control.py
kelvan: as soon as we have more examples we probably should make the connection args reusable, or add an example command for uv with subcommands for the examples

gardena_smart_local_api/examples/app.py
```diff
@@ -0,0 +1,117 @@
...
+            pass
+        return None
+
+    async def send_request(self, msgs: EgressMessageList) -> IngressMessageList | None:
+        await self.send(str(msgs))
+
+        request_ids = {m.request_id for m in msgs}
+
+        for _ in range(10):
+            reply_json = await self.receive()
```
easybe, line 97: I plan to look into adding supports for events as well. So this will likely change in the future.

gardena_smart_local_api/examples/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .app import ExampleApp
+
+__all__ = [ExampleApp]
```
kelvan:
```suggestion
__all__ = ["ExampleApp"]
```

gardena_smart_local_api/examples/app.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        parser.add_argument(
+            "-d",
+            "--device-id",
+            required=False,
+            help='ID of device to control (list IDs with "list" command)',
+        )
+
+        for arg_info in extra_args:
+            name_or_flags = arg_info.pop("name_or_flags")
```
kelvan: this mutates the content of the dicts also outside the constructor, consider making a (deep) copy

gardena_smart_local_api/examples/app.py
```diff
@@ -0,0 +1,117 @@
...
+
+        self.ws = await websockets.connect(
+            uri,
+            ssl=ssl_context,
+            additional_headers=headers,
+            close_timeout=0,
+        )
+
+    async def disconnect(self):
+        if self.ws:
```
kelvan, line 73: `self.ws` is never initialized in `__init__`, so `disconnect()` will raise `AttributeError` if called before `connect()`.
Consider initializing `self.ws = None` in `__init__`.

gardena_smart_local_api/examples/app.py
```diff
@@ -0,0 +1,117 @@
...
+        await self.send(str(msgs))
+
+        request_ids = {m.request_id for m in msgs}
+
+        for _ in range(10):
+            reply_json = await self.receive()
+            if reply_json is None:
+                continue
+            replies = IngressMessageList.model_validate_json(str(reply_json))
+            if not isinstance(replies[0], Reply):
```
kelvan:
```suggestion
            if not replies or not isinstance(replies[0], Reply):
```

gardena_smart_local_api/examples/valve.py
```diff
@@ -0,0 +1,67 @@
...
+                wc = app.devices[app.args.device_id]
+                if not isinstance(wc, WaterControl):
+                    print("Incompatible device selected")
+                    return 1
+                request = wc.build_stop_watering_obj()
+                result = await app.send_request(request)
+                if result is not None and not result[0].success:
+                    print("Failed to stop watering")
+                    if isinstance(result[0], ErrorMessage):
+                        print(f"Error: {result[0].error_message}")
```
kelvan, line 62: there should be a return 1 too I guess?

kelvan (changes requested):

kelvan (approved):
👍

PR #8, opened 2026-03-19
Add PyPI publish workflow

kelvan:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,39 @@
+name: Publish to PyPI
+
+on:
+  release:
+    types: [published]
```
easybe: I would trigger this on Git tag push, no?


```suggestion
  push:
    tags:
      - "v*.*.*"
```

Maybe also consider using https://github.com/softprops/action-gh-release (or similar)?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,57 @@
...
+    permissions:
+      contents: write
+    steps:
+      - name: Download dist artifacts
+        uses: actions/download-artifact@v8
+        with:
+          name: dist
+          path: dist/
+
+      - name: Create GitHub Release
```
easybe:
```suggestion
      - name: Create GitHub release
```

easybe (approved):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,57 @@
...
+        uses: softprops/action-gh-release@v2.6.1
+        with:
+          files: dist/*
```
easybe, line 41:
```suggestion
          files: dist/*
          generate_release_notes: true
```

easybe (commented):
I like this feature. Why not test this already?

PR #10, opened 2026-03-20
Add package build test CI job and hatchling build backend

kelvan:
Switches the build backend from the implicit setuptools default to hatchling, which automatically includes all package files (including `schema/device_models.yaml`) in the wheel without requiring a `MANIFEST.in`.

Also adds a `package` CI job that builds the wheel and runs the test suite against the installed package (not the source tree), catching packaging issues like missing files early.

**Changes:**
- Add `hatchling` as the explicit build backend in `pyproject.toml`
- Fix schema file not being included in the built wheel
- Add `package` CI job: builds wheel, installs into isolated temp environment, runs tests with `--import-mode=importlib`
- Bump version to `0.0.3`


pyproject.toml
```diff
@@ -1,6 +1,13 @@
...
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.build.targets.wheel]
+packages = ["gardena_smart_local_api"]
+
 [project]
 name = "gardena-smart-local-api"
-version = "0.0.2"
+version = "0.0.3"
```
easybe, line 10: Can we merge https://github.com/cloudless-garden/gardena-smart-local-api/pull/7 before bumping the version?

pyproject.toml
```diff
@@ -1,6 +1,13 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
easybe, line 3: I understand why this is better

.github/workflows/ci.yml
```diff
@@ -52,6 +52,26 @@ jobs:
...
+        run: uv build --wheel
+
+      - name: Install wheel and test dependencies
+        run: |
+          UV_PROJECT_ENVIRONMENT=$RUNNER_TEMP/.venv uv sync --group test --no-install-project
+          uv pip install --python $RUNNER_TEMP/.venv dist/*.whl
+
+      - name: Run tests against installed package
+        working-directory: ${{ runner.temp }}
+        run: uv run pytest $GITHUB_WORKSPACE/tests/ --import-mode=importlib
```
easybe, line 72: I don't quite understand. Is the goal to test the package we will distribute? We are testing against the wheel (which we don't release)?, that does not seem right to me.

easybe:
> Add package build test CI job and hatchling build backend

Whenever you need a "and" in your commit message subject, you probably could have created 2 commits.
Alternatively, put the main intent of the commit in the subject and the additional information in the body (in full sentences).

.github/workflows/ci.yml
kelvan: we test against the built package here, this would have e.g. catched the missing yaml file

easybe (approved):
OK, now I understand. The new commits make it much easier to understand what is going on, thanks you very much!

PR #12, opened 2026-03-21
Add Sensor device support (model 18845)

kelvan:
- Add model 18845 to device_models.yaml with full lemonbeat object definitions
- Add Sensor class to gen1.py with properties for soil humidity/temperature, ambient temperature, light, frost warning, and error
- Register Sensor in device_builder and export from devices package
- Add examples/sensor.py demonstrating sensor readings

### DIFF
--- a/gardena_smart_local_api/devices/__init__.py
+++ b/gardena_smart_local_api/devices/__init__.py
@@ -5,12 +5,14 @@ from .device_builder import (
 )
 from .irrigation import Gen1WaterControl
 from .mowers import Gen1Mower
+from .sensors import Sensor1
 
 __all__ = [
     "Device",
     "DeviceMap",
     "Gen1Mower",
     "Gen1WaterControl",
+    "Sensor1",
     "build_discovery_obj",
     "create_devices_from_json",
     "create_devices_from_messages",
--- a/gardena_smart_local_api/devices/sensors.py
+++ b/gardena_smart_local_api/devices/sensors.py
@@ -1,0 +1,79 @@
+from ..messages import EgressMessageList
+from ..resources import IpsoPath
+from .gen1 import Gen1BatteryPoweredGen1Device
+
+
+class Sensor1(Gen1BatteryPoweredGen1Device):
+    @property
+    def temperature(self) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="ambient_temperature",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
+
+    def build_reload_temperature_obj(self) -> EgressMessageList:
+        return self.build_command_obj(self.get_command("measure_ambient_temperature"))
+
+    @property
+    def light(self) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="light",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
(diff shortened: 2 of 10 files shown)


gardena_smart_local_api/examples/sensor.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
```
easybe, line 1: I would rather want a `sensor.py` that works for both sensors.

gardena_smart_local_api/devices/sensors.py
```diff
@@ -0,0 +1,66 @@
...
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="ambient_temperature",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
+
+    def build_reload_temperature_obj(self) -> EgressMessageList:
```
easybe, line 20:
```suggestion
    def build_update_temperature_obj(self) -> EgressMessageList:
```
or
```suggestion
    def build_refresh_temperature_obj(self) -> EgressMessageList:
```
?

easybe (changes requested):
The README needs updating. In that table I would rather want to see the marketing name of the product and the article number. Maybe in the same style as for the migration.

gardena_smart_local_api/devices/sensors.py
kelvan: Update sounds like writing a temperature to the device, refresh could work

gardena_smart_local_api/examples/sensor.py
```diff
@@ -0,0 +1,44 @@
...
+            case "read":
+                if app.args.device_id is None:
+                    print("No device ID provided")
+                    return 1
+                sensor = app.devices[app.args.device_id]
+                if not isinstance(sensor, Sensor):
+                    print("Incompatible device selected")
+                    return 1
+                print(f"Sensor {sensor.id}:")
+                print(f"  Temperature:     {sensor.temperature}°C")
```
easybe, line 34: Soil moisture is missing.

gardena_smart_local_api/devices/sensors.py
kelvan: or maybe use measure? that would it make it clear that it triggers a measurement at the sensor?

easybe: I agree, however, IMHO we should align it with methods like `build_refresh_online_status_obj` (which I find better than  what we have now, `build_reload_online_status_obj`).

kelvan: I think there are 3 different kinds of "reading" a value, getting last announced (property), reloading/refreshing a value from the device and triggering a real remeasurement (that can take some time)
I'll put it on the sync agenda to discuss

easybe: We can leave it for now, as it is consistent. However, we should revise the naming throughout the lib.

gardena_smart_local_api/devices/__init__.py
```diff
@@ -5,12 +5,14 @@
...
 from .irrigation import Gen1WaterControl
 from .mowers import Gen1Mower
+from .sensors import Sensor
 
 __all__ = [
     "Device",
     "DeviceMap",
     "Gen1Mower",
     "Gen1WaterControl",
+    "Sensor",
```
easybe:
```suggestion
    "Sensor1",
```
?

easybe (approved):

easybe (approved):

PR #13, opened 2026-03-21
Add PowerAdapter device support (model 35279)

kelvan:
- Add model 35279 schema with full lemonbeat object definitions
- Add `PowerAdapter` device class with `power_timer`, `is_output_enabled`, `error` properties and `build_enable_output_obj`, `build_disable_output_obj`, `build_identify_obj` methods
- Add `IdentifyMixin` to `Gen1WaterControl` and `_Sensor` (covers `Sensor1`, `Sensor2`)
- Register `PowerAdapter` in device builder and export from devices package
- Add `examples/power.py` demonstrating timed power control
- Add tests with real device data

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ Enables controlling and monitoring GARDENA smart devices in the local network, w
 | smart Water Control           | 19033-20                               | 2812               |
 | smart Dual Water Control      | 19034-20                               | 2814               |
 | smart Pipeline Water Control  | 19050-20                               | 2826               |
+| smart Power Adapter           | 19095-20                               | 35279              |
 | smart SILENO                  | 19060-20, 19060-60                     | 6146               |
 | smart SILENO+                 | 19061-20, 19061-60, 19064-60, 19065-60 | 6146               |
 | smart SILENO city             | 19066-20, 19069-20                     | 29694              |
--- a/gardena_smart_local_api/devices/device_builder.py
+++ b/gardena_smart_local_api/devices/device_builder.py
@@ -4,6 +4,7 @@ from ..messages import IngressMessageList, Reply
 from .device import Device, DeviceMap
 from .irrigation import Gen1WaterControl, Gen2WaterControl
 from .mowers import Gen1Mower1, Gen1Mower2
+from .power import PowerAdapter
 from .sensors import Sensor1, Sensor2
 
 MODEL_NUMBER_MAP: dict[str, type[Device]] = {
@@ -14,6 +15,7 @@ MODEL_NUMBER_MAP: dict[str, type[Device]] = {
     "2814": Gen2WaterControl,
     "2826": Gen2WaterControl,
     "29694": Gen1Mower1,
+    "35279": PowerAdapter,
     "53988": Gen1Mower2,
     "6146": Gen1Mower1,
 }
--- a/gardena_smart_local_api/devices/irrigation.py
+++ b/gardena_smart_local_api/devices/irrigation.py
@@ -1,13 +1,13 @@
 from ..messages import EgressMessageList
 from ..resources import IpsoPath
-from .gen1 import Gen1BatteryPoweredDevice
+from .gen1 import Gen1BatteryPoweredDevice, IdentifyMixin
 from .gen2 import Gen2BatteryPoweredDevice
 
 # Used to indicate that the action was initiated through WebSocket API.
 COMMAND_SOURCE = "18"
 
 
-class Gen1WaterControl(Gen1BatteryPoweredDevice):
+class Gen1WaterControl(IdentifyMixin, Gen1BatteryPoweredDevice):
     def build_set_watering_timer_obj(self, seconds: int) -> EgressMessageList:
         """Set the watering timer.
 
--- a/gardena_smart_local_api/devices/power.py
+++ b/gardena_smart_local_api/devices/power.py
@@ -1,0 +1,37 @@
+from ..messages import EgressMessageList
+from ..resources import IpsoPath
+from .gen1 import Gen1Device, IdentifyMixin
+
+
+class PowerAdapter(IdentifyMixin, Gen1Device):
+    @property
+    def power_timer(self) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="power_timer",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
+
+    @property
+    def is_output_enabled(self) -> bool | None:
+        if (value := self.power_timer) is not None:
+            return value != 0
+        return None
+
+    def build_enable_output_obj(self, seconds: int) -> EgressMessageList:
+        return self.build_write_value_obj(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="power_timer",
+            ),
+            seconds,
+        )
+
+    def build_disable_output_obj(self) -> EgressMessageList:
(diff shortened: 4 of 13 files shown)


gardena_smart_local_api/devices/power.py
```diff
@@ -0,0 +1,54 @@
...
+        return None
+
+    @property
+    def is_on(self) -> bool | None:
+        value = self.power_timer
+        if value is not None:
+            return value != 0
+        return None
+
+    def build_set_power_timer_obj(self, seconds: int) -> EgressMessageList:
```
easybe:
```suggestion
    def build_power_on_obj(self, seconds: int) -> EgressMessageList:
```
or maybe something like
```suggestion
    def build_enable_output_obj(self, seconds: int) -> EgressMessageList:
```
as we are not powering on the device.

gardena_smart_local_api/devices/device_builder.py
```diff
@@ -4,15 +4,17 @@
...
 from .irrigation import Gen1WaterControl, Gen2WaterControl
 from .mowers import Gen1Mower1, Gen1Mower2
+from .power import PowerAdapter
 from .sensors import Sensor1, Sensor2
 
 MODEL_NUMBER_MAP: dict[str, type[Device]] = {
     "18845": Sensor1,
-    "18869": Gen1WaterControl,
     "19040": Sensor2,
+    "18869": Gen1WaterControl,
```
easybe, line 12: I would just leave this sorted alphabetically

gardena_smart_local_api/devices/power.py
```diff
@@ -0,0 +1,51 @@
...
+                object_instance_id="0",
+                resource_name="power_timer",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
+
+    @property
+    def is_on(self) -> bool | None:
```
easybe:
```suggestion
    def is_output_enabled(self) -> bool | None:
```

gardena_smart_local_api/devices/power.py
```diff
@@ -0,0 +1,51 @@
...
+        value = self.power_timer
+        if value is not None:
```
easybe:
```suggestion
        if value := self.power_timer is not None:
```

gardena_smart_local_api/devices/power.py
```diff
@@ -0,0 +1,51 @@
...
+                resource_name="power_timer",
+            ),
+            seconds,
+        )
+
+    def build_disable_output_obj(self) -> EgressMessageList:
+        return self.build_enable_output_obj(0)
+
+    @property
+    def error(self) -> int | None:
```
easybe: This is implemented in `Gen1Device`

gardena_smart_local_api/schema/35279_power.yaml
```diff
@@ -0,0 +1,206 @@
+name: Power
```
easybe:
```suggestion
name: smart Power Adapter
```
IMHO we should use the product name here.

gardena_smart_local_api/schema/35279_power.yaml
```diff
@@ -0,0 +1,206 @@
...
+  lemonbeat:
+    resources:
+      action_paused_until_1:
+        type: vo
+        access: rw
+        persistent: true
+        encoding: hex
+        report_on_update: true
+        description: 'Timestamp until the schedule is paused (Format: YYYYMMDDHHmm, local time)'
+      be_decision_time:
```
easybe: We didn't include this for other devices, so I would not add it here either.

README.md
```diff
@@ -12,6 +12,7 @@ Enables controlling and monitoring GARDENA smart devices in the local network, w
 | smart Water Control           | 19033-20                               | 2812               |
 | smart Dual Water Control      | 19034-20                               | 2814               |
 | smart Pipeline Water Control  | 19050-20                               | 2826               |
+| smart Power                   | 19095-20                               | 35279              |
```
```suggestion
| smart Power Adapter                   | 19095-20                               | 35279              |
```

easybe (changes requested):

gardena_smart_local_api/devices/irrigation.py
```diff
@@ -1,13 +1,13 @@
...
-from .gen1 import Gen1BatteryPoweredDevice
+from .gen1 import Gen1BatteryPoweredDevice, IdentifyMixin
 from .gen2 import Gen2BatteryPoweredDevice
 
 # Used to indicate that the action was initiated through WebSocket API.
 COMMAND_SOURCE = "18"
 
 
-class Gen1WaterControl(Gen1BatteryPoweredDevice):
+class Gen1WaterControl(IdentifyMixin, Gen1BatteryPoweredDevice):
```
easybe, line 10: This change is not related, so, at least should go into a separate commit. What about other device?

gardena_smart_local_api/schema/35279_power.yaml
```diff
@@ -0,0 +1,206 @@
...
+        description: Power timer (write >0=high prio, <0=low prio, 0=stop; read >0=high running, <0=low running, 2^24=infinite on)
+      rf_link_quality:
+        type: vi
+        access: r
+        unit: '%'
+        min: 0
+        max: 100
+        report_on_update: true
+        description: RF link quality percentage
+      rf_link_state:
```
easybe: AFAICT `rf_link_state` is no longer in use and was removed (for all devices).

gardena_smart_local_api/devices/power.py
kelvan: changed it to (value := self.power_timer) otherwise power_timer = 0 would be True as it is not None

gardena_smart_local_api/devices/device_builder.py
kelvan: imo it should be sorted by model number (as a number) or alphabetically by the class name, device classes should be next to each other, maybe also add section comments as the list getting long

kelvan: reverted the change for now

easybe: We can do that. But then there should be a commit that sorts it and explains in the commit message how we should sort the list from now on. I just sorted alphabetically with my editor up until now.

gardena_smart_local_api/examples/power.py
```diff
@@ -0,0 +1,84 @@
...
+            "default": 60,
+            "type": int,
+            "help": "Duration to power on in seconds (default: 60s)",
+        },
+    ]
+
+    async with ExampleApp(extra_args) as app:
+        match app.args.command[0]:
+            case "list":
+                for dev_id, device in app.devices.items():
```
easybe: Can we merge https://github.com/cloudless-garden/gardena-smart-local-api/pull/33 first and use the new method?

easybe (approved):

gardena_smart_local_api/examples/power.py
```diff
@@ -0,0 +1,85 @@
...
+COMPATIBLE = (PowerAdapter,)
+
+
+async def main():
+    extra_args = [
+        {
+            "name_or_flags": ["command"],
+            "nargs": 1,
+            "choices": ("list", "on", "off", "identify"),
+            "help": "List devices or turn power on/off/identify",
```
easybe:
```suggestion
            "help": "List applicable devices, identify or turn power on/off",
```

easybe (changes requested):

easybe (approved):

easybe (changes requested):
see updated previous comment

easybe (approved):

PR #16, opened 2026-03-23
Add Sensor2 device (model 19040)

kelvan:
## Summary

- Adds model `19040` (Sensor II) schema with `gen1` protocol and full command/object definitions
- Adds `Sensor2` class in `devices/sensor.py` exposing `soil_moisture`, `temperature`, `has_frost_warning`, `error`, `battery_level`, `rf_link_quality`
- Registers `Sensor2` in `device_builder` (model number `19040`) and exports it from `__init__`
- Adds `examples/sensor2.py` usage example

## Test plan

- [ ] Connect to a Sensor II device and verify discovery returns a `Sensor2` instance
- [ ] Check all properties return expected values
- [ ] Run `uv run pytest`

### DIFF
--- a/gardena_smart_local_api/devices/sensors.py
+++ b/gardena_smart_local_api/devices/sensors.py
@@ -3,7 +3,25 @@ from ..resources import IpsoPath
 from .gen1 import Gen1BatteryPoweredDevice
 
 
-class Sensor1(Gen1BatteryPoweredDevice):
+class _Sensor(Gen1BatteryPoweredDevice):
+    @property
+    def has_frost_warning(self) -> bool | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="frost_warning",
+            )
+        )
+        if isinstance(value, int):
+            return bool(value)
+        return None
+
+    def build_refresh_soil_moisture_obj(self) -> EgressMessageList:
+        return self.build_command_obj(self.get_command("measure_soil_moisture"))
+
+
+class Sensor1(_Sensor):
     @property
     def temperature(self) -> int | None:
         value = self.get_value(
@@ -16,25 +34,6 @@ class Sensor1(Gen1BatteryPoweredDevice):
         if isinstance(value, int):
             return value
         return None
-
-    def build_refresh_temperature_obj(self) -> EgressMessageList:
-        return self.build_command_obj(self.get_command("measure_ambient_temperature"))
-
-    @property
-    def light(self) -> int | None:
-        value = self.get_value(
-            IpsoPath(
-                object_name="lemonbeat",
-                object_instance_id="0",
-                resource_name="light",
-            )
-        )
-        if isinstance(value, int):
-            return value
-        return None
-
-    def build_refresh_light_obj(self) -> EgressMessageList:
-        return self.build_command_obj(self.get_command("measure_light"))
 
     @property
     def soil_moisture(self) -> int | None:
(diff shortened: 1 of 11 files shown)


gardena_smart_local_api/devices/sensor.py
```diff
@@ -0,0 +1,66 @@
+from ..messages import EgressMessageList
```
easybe, line 1: This should be in `sensors.py`

gardena_smart_local_api/devices/sensor.py
```diff
@@ -0,0 +1,66 @@
...
+from ..resources import IpsoPath
+from .gen1 import Gen1BatteryPoweredGen1Device
+
+
+class Sensor2(Gen1BatteryPoweredGen1Device):
```
easybe: How about only having one `Sensor` class that works for both sensors?

easybe (changes requested):
See comments in #12

gardena_smart_local_api/devices/sensor.py
kelvan: sensor 2 has no light sensor, so there is no clean way to combine it imo

easybe: I would consider returning None. But yes, maybe it is better to have distinct types.

kelvan: imho having a property that always return None is not clean code
this would also lead to an entity in HA that is always unknown, or need some ugly hacks
but we could think about starting to use protocols, or create a Sensor class that both inherit from (but than it starts to get deep)

kelvan:
this need overhaul after #12 done

gardena_smart_local_api/devices/sensors.py
```diff
@@ -77,3 +77,66 @@ def error(self) -> int | None:
         if isinstance(value, int):
             return value
         return None
+
+
+class Sensor2(Gen1BatteryPoweredGen1Device):
```
easybe: Needs rebasing

gardena_smart_local_api/devices/sensors.py
```diff
@@ -77,3 +77,66 @@ def error(self) -> int | None:
...
+                object_instance_id="0",
+                resource_name="frost_warning",
+            )
+        )
+        if isinstance(value, int):
+            return bool(value)
+        return None
+
+    @property
+    def error(self) -> int | None:
```
easybe: This is a common gen1 property and should be moved to `Gen1Device`.

gardena_smart_local_api/devices/sensors.py
```diff
@@ -77,3 +77,66 @@ def error(self) -> int | None:
...
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="error",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
+
+    def build_refresh_soil_moisture_obj(self) -> EgressMessageList:
```
easybe: Common sensor-related methods I would move to a `_Sensor` base class.

easybe (changes requested):

gardena_smart_local_api/devices/sensors.py
kelvan: Sensor1 uses `soil_humidity` instead of `soil_moisture`, `has_frost_warning` and `build_refresh_soil_moisture_obj` (firmware seems to be inconsistent here) is the same, combined them

gardena_smart_local_api/devices/sensors.py
```diff
@@ -1,41 +1,42 @@
+from abc import ABC
+
 from ..messages import EgressMessageList
 from ..resources import IpsoPath
 from .gen1 import Gen1BatteryPoweredDevice
 
 
-class Sensor1(Gen1BatteryPoweredDevice):
+class _Sensor(Gen1BatteryPoweredDevice, ABC):
```
easybe: I considered inheriting from `ABC` too but then decided not to, as:

> ### When ABC is the right choice
>
> * You want to enforce that subclasses implement specific methods (the main value of `@abstractmethod`)
> * You want tooling/IDEs to understand the interface and flag missing implementations
> * You're designing a formal interface/protocol
>
> ### When ABC is overkill
>
> * You just want to prevent direct instantiation with no interface contract
> * It's a simple internal class — a docstring convention or __new__ guard is enough
> * You're using Python's Protocol for structural subtyping instead
>
>The real power of ABC isn't preventing instantiation — it's enforcing an interface contract on subclasses. Preventing instantiation is just a side effect of having unimplemented abstract methods.

easybe (approved):

PR #17, opened 2026-03-24
Add repository URL to pyproject.toml for PyPI

easybe:
This is in line with:
https://packaging.python.org/en/latest/specifications/well-known-project-urls/

---

Tested here: https://test.pypi.org/project/gardena-smart-local-api/0.0.4.dev1/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,9 @@ dependencies = [
     "pyyaml>=6.0",
 ]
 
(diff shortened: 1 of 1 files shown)


easybe:
gitlint failing in this situation is fine. As there are exceptions to the rules, I made the check optional for merge.

kelvan (approved):

PR #24, opened 2026-03-27
Add gen2 Water Control support

easybe:
Supported devices:

* smart Water Control (Article No. 19033-20)
* smart Dual Water Control (Article No. 19034-20)
* smart Pipeline Water Control (Article No. 19050-20)

### DIFF
--- a/gardena_smart_local_api/devices/device.py
+++ b/gardena_smart_local_api/devices/device.py
@@ -153,6 +153,17 @@ class Device(BaseModel):
         """
         return self.objects.get(object_name, {}).get(instance_id)
 
+    def get_object_instance_ids(self, object_name: str) -> list[str]:
+        """Get list of all instance IDs for an object.
+
+        Args:
+            object_name: Name of the object
+
+        Returns:
+            List of instance ID strings
+        """
+        return [k for k in self.objects.get(object_name, {}).keys() if k != "_urn"]
+
     def get_resource(
         self, object_name: str, instance_id: str | int, resource_name: str
     ) -> Any | None:
@@ -207,17 +218,6 @@ class Device(BaseModel):
                 )
             ]
         )
-
-    def list_instances(self, object_name: str) -> list[str]:
-        """Get list of all instance IDs for an object.
-
-        Args:
-            object_name: Name of the object
-
-        Returns:
-            List of instance ID strings
-        """
-        return list(self.objects.get(object_name, {}).keys())
 
     @property
     def is_online(self) -> bool | None:
--- a/gardena_smart_local_api/devices/device.py
+++ b/gardena_smart_local_api/devices/device.py
@@ -1,6 +1,6 @@
 from collections.abc import MutableMapping
 from functools import cached_property
-from typing import Any, ClassVar
+from typing import Any, ClassVar, cast
 
 from pydantic import BaseModel, Field, RootModel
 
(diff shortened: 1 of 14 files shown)


gardena_smart_local_api/devices/device_builder.py
```diff
@@ -2,12 +2,13 @@
...
 from ..messages import IngressMessageList, Reply
 from .device import Device, DeviceMap
-from .irrigation import Gen1WaterControl
+from .irrigation import Gen1WaterControl, Gen2WaterControl
 from .mowers import Gen1Mower
 
 MODEL_NUMBER_MAP: dict[str, type[Device]] = {
     "18869": Gen1WaterControl,
     "53988": Gen1Mower,
+    "2814": Gen2WaterControl,
```
kelvan, line 14: there are 3 Gen2 WaterControls, this is the Dual, we should talk about how we want to do the inheritance/protocols for similar devices

easybe: Yes. Will add the others later. Same class can be used for all 3.

gardena_smart_local_api/schema/device_models.yaml
```diff
@@ -855,3 +855,714 @@ models:
             min: 0
             max: 255
             description: Message type
+  '2812':
+    name: smart Water Control
+    type: Water Control
```
easybe: Do we need this? `device_type` is sent by the device.

kelvan: I think we can remove it and use the info from the device where needed

gardena_smart_local_api/devices/irrigation.py
```diff
@@ -1,6 +1,9 @@
 from ..messages import EgressMessageList
 from ..resources import IpsoPath
 from .gen1 import Gen1BatteryPoweredDevice
+from .gen2 import Gen2BatteryPoweredDevice
+
+COMMAND_SOURCE = "18"
```
kelvan, line 7: what does the command source represent, what does 18 mean?

gardena_smart_local_api/examples/irrigation.py
```diff
@@ -50,12 +76,32 @@ async def main():
...
                     print("Incompatible device selected")
                     return 1
-                request = wc.build_stop_watering_obj()
+
+                if isinstance(wc, Gen1WaterControl):
+                    request = wc.build_stop_watering_obj()
+                else:
+                    if (
+                        app.args.valve_id is not None
+                        and app.args.valve_id >= wc.valve_count
```
kelvan, line 91: does not catch negative numbers

gardena_smart_local_api/resources.py
```diff
@@ -29,7 +32,49 @@ def _parse_opaque(v: object) -> bytes | None:
...
+                return matches
+        return cast(list[str], v)
+    raise ValueError(f"Expected list[str] or str, got: {type(v)}")
+
+
+StringArray = Annotated[
+    list[str],
+    BeforeValidator(_parse_as),
+    PlainSerializer(
+        lambda v: [",".join(f"{i}='{s}'" for i, s in enumerate(v))],
```
kelvan, line 72: what happens if s includes single-quotes?

gardena_smart_local_api/devices/device.py
```diff
@@ -119,6 +124,20 @@ def build_read_value_obj(self, path: IpsoPath) -> EgressMessageList:
             ]
         )
 
+    def build_execute_obj(
```
kelvan, line 127: should have an access check (`access: x` in model definition) like the read/write checks

kelvan (changes requested):
needs some basic tests for the new devices

gardena_smart_local_api/devices/irrigation.py
easybe:
```
Start action of this actuator and create a new timeslot corresponding to this action. Takes two arguments:
  0: integer: start source. The only purpose is to be copied over to corresponding timeslot.
     You can see a list of values in timeslot/initial_source.
  1: integer: duration in seconds.
```
[source](https://dev.azure.com/HQV-Gardena/SG-Gateway/_git/sg-bnw-ipso-registry?path=/definitions/actuator.xml&version=GBar/update-descriptions&line=132&lineEnd=133&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

```
The reason why this timeslot was created, specifically why `start` has exactly this value. Also applies to why `adjusted_stop` has its value in case `adjustment_source` is NONE.

Values
------
0: NONE: Only relevant for adjustment.
1: PHYSICAL_BUTTON: Relevant for start reason only.
2: ERROR: Relevant for stop reason only, e.g. valve overcurrent.
3: SCHEDULE: For scheduled events.
4: BOOK_KEEPING: Only used internally by the scheduler.
10: BLUETOOTH: Used by Bluetooth application interface.
16: SMART_SYSTEM: Used by smart system app, e.g. for manual start & stop actions.
17: HOMEKIT: Used by HomeKit a.k.a. Apple Home
18: WEB_SOCKET: Used by websocketd (for integrations like Home Assistant).
```
[source](https://dev.azure.com/HQV-Gardena/SG-Gateway/_git/sg-bnw-ipso-registry?path=/definitions/timeslot.xml&version=GBar/update-descriptions&line=83&lineEnd=84&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

gardena_smart_local_api/schema/device_models.yaml
kelvan: created #28 removing type

gardena_smart_local_api/devices/irrigation.py
kelvan: maybe add a short comment in the code explaining it

gardena_smart_local_api/resources.py
easybe: Bad things. I added a comment.

easybe:
> needs some basic tests for the new devices

I will do that at a later time

kelvan (approved):

PR #26, opened 2026-03-30
Rename build_reload_* methods to build_refresh_*

kelvan:
Renames all `build_reload_*` methods to `build_refresh_*`.


easybe (approved):
> Renames all build_reload_* methods to build_refresh_* for consistency.

We don't use `refresh` yet, so technically consistency is not the issue.

PR #27, opened 2026-03-31
Add support for remaining gen1 mowers

easybe:
Support for smart SILENO(+) and smart SILENO city/life, without the LONA feature, were still missing.

The signature of `Gen1Mower2.build_stop_mowing_obj()` is changed to be more in line with the corresponding `Gen1Mower1` method.

---

Needs testing with older models.

Run the example code as follows:
```
uv run gardena_smart_local_api/examples/mower.py -g $GW -p $PW list
uv run gardena_smart_local_api/examples/mower.py -g $GW -p $PW -d $SGTIN start $DURATION
uv run gardena_smart_local_api/examples/mower.py -g $GW -p $PW -d $SGTIN stop
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,20 @@ Enables controlling and monitoring GARDENA smart devices in the local network, w
 
 ## Supported Devices
 
-| Device                       | Article Number | Model (EAN suffix) |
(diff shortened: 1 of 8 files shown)


kelvan (approved):
lgtm

PR #28, opened 2026-03-31
Remove type field from device models

kelvan:
## Summary

- Remove the `type` field from `BaseModelDefinition` as the type is reported by the devices themselves and doesn't need to be hardcoded in the YAML schema
- Replace all usages of `model_definition.type` with `model_definition.name` across examples and repr methods
- Remove the corresponding `type` assertions from tests

## Test plan

- [x] All 60 existing tests pass (`uv run pytest`)
- [x] `ruff check`, `ruff format`, and `ty check` all pass


easybe (approved):
Thank you very much!

PR #33, opened 2026-04-02
Improve examples

easybe:
To watch the sensor values for 60 seconds, run:

    uv run gardena_smart_local_api/examples/sensor.py -g $GW -p $PW \
        -d $DEVICE_ID read 60

### DIFF
--- a/gardena_smart_local_api/examples/app.py
+++ b/gardena_smart_local_api/examples/app.py
@@ -8,6 +8,7 @@ from typing import Self
 import websockets
 
 from gardena_smart_local_api.devices import (
+    Device,
     DeviceMap,
     build_discovery_obj,
     create_devices_from_messages,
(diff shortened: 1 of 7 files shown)


gardena_smart_local_api/examples/app.py
```diff
@@ -83,27 +114,27 @@ async def receive(self) -> str | None:
...
-            if reply_json is None:
-                continue
-            replies = IngressMessageList.model_validate_json(reply_json)
-            if not replies or not isinstance(replies[0], Reply):
-                continue
-            if {r.request_id for r in replies} == request_ids:
-                return replies
+            for request_id in request_ids:
+                if request_id in self.replies:
+                    reps.append(self.replies[request_id])
```
kelvan: if there a two requests and one sends a report, this will readd the same reply in each iteration, after the second iteration it will break and return two copies of the same response

kelvan (changes requested):

gardena_smart_local_api/examples/app.py
easybe: You were too fast, I fixed it in the meantime.

kelvan (approved):

PR #35, opened 2026-04-02
Fix blocking scandir call in async event loop

kelvan:
## Summary

- `ModelLoader.load()` called `Path.glob()` synchronously, which triggers `os.scandir` and blocks the Home Assistant event loop
- Wrapped the glob call in `asyncio.to_thread()` to offload it to a thread pool executor

## Test plan

- [ ] Deploy to Home Assistant and confirm the blocking call warning no longer appears in logs


easybe (approved):
LGTM

PR #37, opened 2026-04-03
Add gen1 mower status property & improve examples

easybe:


gardena_smart_local_api/devices/mowers.py
```diff
@@ -1,11 +1,51 @@
...
+    def status(self) -> Gen1MowerStatus | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="status",
+            )
+        )
+        if isinstance(value, int):
+            return Gen1MowerStatus(value)
```
kelvan: if value is not a known status code this will raise an ValueError, maybe add an UNKNOWN status or return None?

kelvan (changes requested):

kelvan (approved):

PR #39, opened 2026-04-05
Add support for gen2 Irrigation Control

easybe:


gardena_smart_local_api/devices/gen2.py
```diff
@@ -1,20 +1,21 @@
...
 from ..model_loader import Gen2ModelDefinition
-from ..resources import IpsoPath
+from ..resources import VALUE_TYPES, IpsoPath
 from .device import Device
 
 
-class Gen2Device(Device):
-    model_definition: Gen2ModelDefinition = Field()
-    service: ClassVar[str] = "lwm2mserver"
+class _DeviceProtocol(Protocol):
```
easybe, line 10: I suggest this naming scheme. Technically the Protocol could be moved to `device.py` but I suggest to keep it private for now.

gardena_smart_local_api/devices/irrigation.py
```diff
@@ -1,12 +1,32 @@
...
 from ..resources import IpsoPath
 from .gen1 import Gen1BatteryPoweredDevice, IdentifyMixin
-from .gen2 import Gen2BatteryPoweredDevice
+from .gen2 import Gen2BatteryMixin, Gen2Device
 
 # Used to indicate that the action was initiated through WebSocket API.
 COMMAND_SOURCE = "18"
 
 
+class TimeslotState(Enum):
```
kelvan, line 12: is timeslot needed without the cloud backend?

easybe: Yes, manual watering also uses a timeslot. Basically representing the timer.

gardena_smart_local_api/devices/irrigation.py
```diff
@@ -163,3 +187,29 @@ def build_close_all_valves_obj(self) -> EgressMessageList:
             ),
             [COMMAND_SOURCE],
         )
+
+    def is_valve_open(self, valve_id: int) -> bool | None:
+        return self.get_timeslot_state(valve_id) == TimeslotState.RUNNING
+
+    def get_timeslot_state(self, timeslot_id: int) -> TimeslotState | None:
```
kelvan, line 194: timeslot_id is the same as valve_id? so timeslot_id for valve_id=1 is also 1 I guess?
should we name it valve_id here too to prevent confusion?

kelvan (approved):
don't forget to add tests later on ;)

gardena_smart_local_api/devices/irrigation.py
easybe: That is how I had in the beging. But as `valve_id` and  `timeslot_id` are only the same when checking the current state (there are also timeslots for schedules) I decided against it.

easybe:
> don't forget to add tests later on ;)

https://github.com/cloudless-garden/gardena-smart-local-api/issues/20

PR #44, opened 2026-04-08
Add error_code and error properties to device classes

kelvan:
## Summary
- Maps device error codes to human-readable strings via enums
- Provides a uniform `error_code` (int) / `error` (str) interface across all Gen1 device classes
- Updates examples to display errors

## Test plan
- [ ] Verify `error` returns `None` when no error is present (code 0)
- [ ] Verify `error` returns the enum name string for known error codes
- [ ] Verify `error` falls back to the raw int string for unknown codes


rettichschnidi:
Will this be worked on, or how about closing this PR?

kelvan:
we should discuss how to handle/show errors, we need a good solution that work for all devices

PR #45, opened 2026-04-08
Add Sphinx documentation with GitHub Pages deployment

kelvan:
## Summary

- Sets up Sphinx docs with the `furo` theme, `sphinx-autoapi` for API reference, and `myst-parser` for Markdown support
- Adds guide pages covering architecture, schema file format, and example scripts
- Moves contributing content from README to `CONTRIBUTING.md`
- Adds a GitHub Actions workflow that builds and deploys docs to GitHub Pages on push to `main`

## Notes

GitHub Pages source must be set to **GitHub Actions** in the repo settings before the deployment workflow can run (Settings → Pages → Source). :heavy_check_mark: 

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,0 +1,90 @@
+# Architecture
+
+## Connection
+
+The **GARDENA smart Gateway** is reachable over a TLS WebSocket on port **8443** and
+acts as a bridge to the individual devices. The library wraps the protocol; WebSocket
+connection handling is left to the caller (e.g. the Home Assistant integration).
+
+Authentication uses HTTP Basic Auth with a password equal to the first 8 characters
+of the ID printed on the back of the gateway (username is ignored on the gateway):
+
+```
+wss://GARDENA-123456:8443
+Authorization: Basic base64("_:a3f2c8b1")
+```
+
+The gateway's TLS certificate is self-signed, so certificate verification must be
+disabled on the client side.
+
+## Protocol
+
+GARDENA devices speak a JSON-based protocol derived from
+[OMA LwM2M](https://www.openmobilealliance.org/solutions/lightweight-m2m-lwm2m/)
+(Lightweight M2M) and the underlying
+[IPSO Smart Objects](https://www.openmobilealliance.org/release/LightweightM2M/)
+resource model.
+
+State is organized as a tree of **objects** and **resources**:
+
+```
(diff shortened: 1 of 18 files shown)


docs/architecture.md
```diff
@@ -0,0 +1,89 @@
+# Architecture
+
+## Connection
+
+The library connects to the **GARDENA Smart Gateway** over a TLS WebSocket on port
```
easybe:
```suggestion
The library connects to the **GARDENA smart Gateway** over a TLS WebSocket on port
```

docs/architecture.md
```diff
@@ -0,0 +1,89 @@
...
+**8443**. The gateway acts as a bridge between the library and the individual devices.
+
+Authentication uses HTTP Basic Auth with a fixed username `_` and a password equal to
```
easybe: Username is actually ignored

docs/architecture.md
```diff
@@ -0,0 +1,89 @@
...
+the first 8 characters of the ID printed on the back of the gateway:
+
+```
+wss://GARDENA-123456:8443
+Authorization: Basic base64("_:MYPASSWD")
```
easybe: MYPASSWD is not a good example as it is does not show that it is derived from the gateway ID.

easybe (changes requested):
IMHO, instead of AI generated documentation (that someone needs to keep up to date and verify it is not hallucinated), we should just use the README as front page of the docs. I did the same [here](https://github.com/UpstreamDataInc/goosebit/tree/master/docs) using symlinks. If we need it to be more elaborate, we extend the README.

Where I see an actual benefit in using Sphinx is the API reference, generated from the source code (not by AI).

kelvan:
> Where I see an actual benefit in using Sphinx is the API reference, generated from the source code (not by AI).

I also want to have the docs as a place to start if you want to create a new lib for e.g. java, how to read the schema files, add icd specs and so on

kelvan:
should we publish it on readthedocs too?

easybe:
> should we publish it on readthedocs too?

I don't know. Seems a bit redundant but might help for SEO.

kelvan:
> > should we publish it on readthedocs too?
> 
> I don't know. Seems a bit redundant but might help for SEO.

better keep it on github pages only for now

rettichschnidi:
@kelvan @easybe How about closing this?

PR #47, opened 2026-04-08
Add status command to irrigation example

easybe:


rettichschnidi:
Is this still relevant/will it be worked on? Suggesting to close otherwise.

PR #51, opened 2026-04-27
Exclusion feature

kelvan:
Send factory reset command to exclude a device (same as the app does)


gardena_smart_local_api/devices/device.py
```diff
@@ -233,6 +233,24 @@ def build_set_resource_obj(
             ]
         )
 
+    def build_exclusion_device_obj(self) -> EgressMessageList:
```
easybe:
```suggestion
    def build_exclusion_obj(self) -> EgressMessageList:
```

gardena_smart_local_api/examples/exclusion.py
```diff
@@ -0,0 +1,54 @@
...
+]
+
+
+async def main():
+    async with ExampleApp((Device,), extra_args) as app:
+        match app.args.command[0]:
+            case "list":
+                app.list_devices()
+
+            case "exclude":
```
easybe: I would merge this with `inclusion.py`. How about `device.py include/exclude`?

gardena_smart_local_api/sgtin96.py
```diff
@@ -21,39 +21,48 @@ def _gtin_checksum(digits: str) -> str:
     return "0" if remainder == 0 else str(10 - remainder)
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
```
easybe: Why do we need this?

easybe (changes requested):

gardena_smart_local_api/examples/exclusion.py
kelvan: Thought of this, was unsure, but probably better yes

gardena_smart_local_api/sgtin96.py
kelvan: we don't, it's over-optimisation I'll change it

easybe:
I've put the file renaming into a separate commit. This makes the example reviewable.

gardena_smart_local_api/sgtin96.py
```diff
@@ -23,37 +23,46 @@ def _gtin_checksum(digits: str) -> str:
 
 @dataclass
 class Sgtin96Info:
```
easybe: As you are changing the interface you might as well rename the class according to PEP 8 (in a separate commit):
```suggestion
class SGTIN96Info:
```

gardena_smart_local_api/sgtin96.py
```diff
@@ -23,37 +23,46 @@ def _gtin_checksum(digits: str) -> str:
...
+    hex_str: str
```
easybe: It would be helpfully if the commit message body would explain why we need this.

kelvan: actually we do not need it anymore, I used it in an earlier version for the HA exclusion feature, not used anymore as far as I can see. need to confirm this

easybe: Let's finish this and then merge https://github.com/cloudless-garden/gardena-smart-local-api/pull/55

easybe (approved):

PR #57, opened 2026-05-08
Added additional sensor entities

nurtext:
New sensor API readouts using mixins for:
https://github.com/cloudless-garden/ha-gardena-smart-local-preview/pull/29

### DIFF
--- a/gardena_smart_local_api/devices/gen1.py
+++ b/gardena_smart_local_api/devices/gen1.py
@@ -107,7 +107,7 @@ class Gen1GuideWireLengthMixin:
         if isinstance(value, int):
             return int(value)
         return None
-    
+
 
 class Gen1IdentifyMixin:
     def build_identify_obj(self: _Gen1DeviceProtocol) -> EgressMessageList:
--- a/gardena_smart_local_api/devices/gen1.py
+++ b/gardena_smart_local_api/devices/gen1.py
@@ -34,6 +34,81 @@ class Gen1BatteryMixin:
         return self.build_command_obj(self.get_command("measure_battery"))
 
 
+class Gen1ChargingCyclesMixin:
+    @property
+    def charging_cycles(self: _Gen1DeviceProtocol) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="charging_cycles",
+            )
+        )
+        if isinstance(value, int):
+            return int(value)
+        return None
+
+
+class Gen1CuttingTimeMixin:
+    @property
+    def cutting_time(self: _Gen1DeviceProtocol) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="cutting_time",
+            )
+        )
+        if isinstance(value, int):
+            return int(value)
+        return None
+
+
+class Gen1RunningTimeMixin:
+    @property
+    def running_time(self: _Gen1DeviceProtocol) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
+                object_instance_id="0",
+                resource_name="running_time",
+            )
+        )
+        if isinstance(value, int):
+            return int(value)
+        return None
+
+
+class Gen1CollisionsMixin:
+    @property
+    def collisions(self: _Gen1DeviceProtocol) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="lemonbeat",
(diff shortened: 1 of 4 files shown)


easybe:
Read https://github.com/cloudless-garden/gardena-smart-local-api#contributing for hints on how to fix the CI issues.

gardena_smart_local_api/devices/gen2.py
```diff
@@ -28,6 +28,66 @@ def battery_level(self: _DeviceProtocol) -> float | None:
         return None
 
 
+class Gen2ChargingCyclesMixin:
```
easybe: We really should not overuse Mixins. We only use them when there is an actual benefit. This for example can go directly into `Gen2Mower`.

easybe (changes requested):

gardena_smart_local_api/devices/gen2.py
nurtext: Sorry, kinda new to Python and it's patterns. Thought this would be the right way. I will continue to rework my code. Guess I should remove all these mixins and add them as properties/defs instead?

easybe: > Thought this would be the right way. 

It actually was in the 00s 🤣 Nowadays, Mixins are somewhat of an anti pattern. But, they still can make sense in some cases 🤷

nurtext: @easybe I removed the mixins I added, but kept the ones that have been commited before. Is that okay?

easybe: 👍 Of course don't just delete stuff, those Mixins are there for a reason.

gardena_smart_local_api/devices/gen1.py
```diff
@@ -99,6 +99,71 @@ def build_command_obj(self, command: int) -> EgressMessageList:
...
+                object_instance_id="0",
+                resource_name="charging_cycles",
+            )
+        )
+        if isinstance(value, int):
+            return int(value)
+        return None
+
+    @property
+    def cutting_time(self: _Gen1DeviceProtocol) -> int | None:
```
easybe, line 116: The docstring should doument the unit (hours) of the returned value.

gardena_smart_local_api/devices/gen2.py
```diff
@@ -31,3 +31,55 @@ def battery_level(self: _DeviceProtocol) -> float | None:
...
+    def cutting_time(self: _DeviceProtocol) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="statistics",
+                object_instance_id="0",
+                resource_name="total_cutting_time",
+            )
+        )
+        if isinstance(value, int):
+            return int(value)
```
easybe, line 58: The value is here in seconds. We should convert it to hours to align with gen1 mowers.

gardena_smart_local_api/examples/mower.py
```diff
@@ -74,7 +74,14 @@ async def main():
...
                     return 1
                 assert isinstance(mower, COMPATIBLE)
-                print(f"Mower state: {mower.state}")
+                print(f"Mower State: {mower.state}")
+                print(f"Battery Level: {mower.battery_level}%")
+                print(f"Charging Cycles: {mower.charging_cycles}")
+                print(f"Cutting Time: {mower.cutting_time}h")
+                print(f"Running Time: {mower.running_time}h")
+                print(f"Collisions: {mower.collisions}")
+                print(f"Guide Wire Length: {mower.guide_wire_length}m")
```
easybe, line 83: Gen2 mowers don't provide this, so you need to check for the type.

gardena_smart_local_api/devices/mowers.py
```diff
@@ -1,8 +1,14 @@
 from ..messages import EgressMessageList
 from ..resources import IpsoPath
 from ._enums import _LowerNameEnum
-from .gen1 import Gen1BatteryMixin, Gen1Device
-from .gen2 import Gen2BatteryMixin, Gen2Device
+from .gen1 import (
+    Gen1BatteryMixin,
```
easybe, line 5: This does not need to be changed.

gardena_smart_local_api/examples/mower.py
```diff
@@ -74,7 +74,14 @@ async def main():
...
+                print(f"RF Link Quality: {mower.rf_link_quality}%")
```
easybe, line 84: We don't support this yet for gen2 devices.

easybe (changes requested):

rettichschnidi:
@nurtext Will you work on this PR? Suggesting to close it otherwise.

PR #59, opened 2026-05-16
Remove checked in file from .gitignore

rettichschnidi:
We have uv.lock checked into Git, and it makes sense to me. We should however remove it from .gitignore in order to avoid confusion.


kelvan (approved):
Good find, especially as we have feature rich example cli apps in the lib the lock file makes sense there, and should be up to date.

PR #60, opened 2026-05-19
Make gen1 mowers report "paused" state

easybe:
Only report "parked" if we are certain that the mower is in the charging station.


kelvan (approved):
Lgtm

FredBit90 (approved):

PR #63, opened 2026-06-12
Replace Pump valve API with start/stop methods

kelvan:
A pump only switches on/off, so only expose dedicated start/stop methods instead of valves.

### DIFF
--- a/gardena_smart_local_api/devices/irrigation.py
+++ b/gardena_smart_local_api/devices/irrigation.py
@@ -271,17 +271,34 @@ class Gen2IrrigationControl(_Gen2Irrigation):
     pass
 
 
-class Pump(Gen1IdentifyMixin, Gen1FrostWarningMixin, _Gen1Irrigation):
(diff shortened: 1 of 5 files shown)


gardena_smart_local_api/devices/irrigation.py
```diff
@@ -271,17 +271,33 @@ class Gen2IrrigationControl(_Gen2Irrigation):
...
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
 
-    @property
-    def valve_ids(self) -> list[int]:
-        return [0]
+    def build_set_watering_timer_obj(
```
easybe: Wouldn't start/stop or enable/disable be a nicer interface? Maybe even align the interface of the Power?

kelvan: yes, start/stop would fit better

easybe (approved):

PR #64, opened 2026-06-15
Water Control duo fixes

kelvan:
- valve open property assumed actuator id == valve id, but this is not the case (for manual triggers)
- objects property is cached as it loads the static schema, but for multi instance elements this lead to valves not showing up in HA integration

### DIFF
--- a/gardena_smart_local_api/devices/irrigation.py
+++ b/gardena_smart_local_api/devices/irrigation.py
@@ -245,7 +245,19 @@ class _Gen2Irrigation(Gen2Device):
     def is_valve_open(self, valve_id: int = 0) -> bool | None:
         if valve_id not in self.valve_ids:
             raise ValueError(f"Invalid valve ID {valve_id}")
-        return self.get_timeslot_state(valve_id) == TimeslotState.RUNNING
+        for timeslot_id in self.get_object_instance_ids("timeslot"):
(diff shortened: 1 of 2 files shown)


easybe (approved):
LGTM

easybe (commented):
Regarding the commit message. I think it should rather be `Timeslot IDs do not necessarily match valve IDs`. And I would explain that this fixes detecting physical button presses.

gardena_smart_local_api/devices/irrigation.py
```diff
@@ -245,7 +245,19 @@ def build_close_all_valves_obj(self) -> EgressMessageList:
     def is_valve_open(self, valve_id: int = 0) -> bool | None:
         if valve_id not in self.valve_ids:
             raise ValueError(f"Invalid valve ID {valve_id}")
-        return self.get_timeslot_state(valve_id) == TimeslotState.RUNNING
+        for timeslot_id in self.get_object_instance_ids("timeslot"):
```
easybe, line 248: See comment regarding commit message

easybe (approved):

easybe:
> Regarding the commit message. I think it should rather be `Timeslot IDs do not necessarily match valve IDs`. And I would explain that this fixes detecting physical button presses.

I was referring to the sentence in the body, not change the title!

PR #65, opened 2026-06-16
Explain debugging on the gateway

rettichschnidi:
This helps to understand what exactly is going on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,70 @@ uv run gardena_smart_local_api/examples/irrigation.py --help
 
 ## Contributing
 
+### Debugging
+
+Various strategies exist to figure out why things are not working the way you want them to. This section lists some
+common ones.
+
+#### Verbose logging on the gateway
+
+websocketd:
+
+```
+mkdir -p /etc/systemd/system/websocketd.service.d
+cat << EOF > /etc/systemd/system/websocketd.service.d/debug.conf
+[Service]
+Environment=RUST_LOG=debug
+EOF
+systemctl daemon-reload && systemctl restart websocketd.service
+```
+
+lemonbeatd:
+
+```
+mkdir -p /etc/systemd/system/lemonbeatd.service.d
+cat << EOF > /etc/systemd/system/lemonbeatd.service.d/debug.conf
+[Service]
+Environment=RUST_LOG=debug
+EOF
+systemctl daemon-reload && systemctl restart lemonbeatd.service
+```
+
+lwm2mserver (add `--log-level=lwm2mserver.event=DEBUG` as a argument to lwm2mserver):
+
+```
+mkdir -p /etc/systemd/system/lwm2mserver.service.d
+cat << EOF > /etc/systemd/system/lwm2mserver.service.d/debug.conf
+[Service]
+ExecStart=
+ExecStart=/usr/bin/lwm2mserver ppp0 --bind-to-device \\
+                                    --server-uri "coap://[fc00::6:100:0:0]" \\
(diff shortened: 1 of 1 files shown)


kelvan (approved):
lgtm

kelvan (approved):
Lgtm

kelvan:
maybe add a note that the cloudadapter is actually handy to keep enabled while developing to log events triggered by the app for some reverse-engineering :)

rettichschnidi:
> maybe add a note that the cloudadapter is actually handy to keep enabled while developing to log events triggered by the app for some reverse-engineering :)

I gave it a try

kelvan (approved):

README.md
```diff
@@ -54,6 +54,67 @@ uv run gardena_smart_local_api/examples/irrigation.py --help
 
 ## Contributing
 
+### Debugging
+
```
easybe, line 58: IMHO there should some text here that explains how this helps.

rettichschnidi: I gave it a try. What do you think?

README.md
```diff
@@ -54,6 +54,70 @@ uv run gardena_smart_local_api/examples/irrigation.py --help
...
+mkdir -p /etc/systemd/system/cloudadapter.service.d
+cat << EOF > /etc/systemd/system/cloudadapter.service.d/debug.conf
+[Service]
+ExecStart=
+ExecStart=/usr/bin/cloudadapter --verbosity --debug
+EOF
+systemctl daemon-reload && systemctl restart cloudadapter.service
+```
+
+Please note: Cloudadapter is not needed for local usage and disabling the service saves resources. However, for reverse-
```
easybe: How about using an "alert" for this? 

https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

README.md
```diff
@@ -54,6 +54,70 @@ uv run gardena_smart_local_api/examples/irrigation.py --help
...
+### Debugging
+
+Various strategies exist to figure out why things are not working the way you want them to. This section lists some
+common ones.
+
+#### Verbose logging on the gateway
+
+websocketd:
+
+```
```
```suggestion
```txt
```

This makes the VS Code linter ([markdownlint](https://github.com/DavidAnson/markdownlint/blob/v0.40.0/doc/md040.md)) happy.

easybe (changes requested):

README.md
rettichschnidi: Done

README.md
rettichschnidi: Sadly, GH renders it weirdly (red):

image

Using shell or console I think would also be wrong, so I leave it the way it was.

easybe (approved):

README.md
```diff
@@ -54,6 +54,70 @@ uv run gardena_smart_local_api/examples/irrigation.py --help
...
+
+cloudadapter:
+
+```
+mkdir -p /etc/systemd/system/cloudadapter.service.d
+cat << EOF > /etc/systemd/system/cloudadapter.service.d/debug.conf
+[Service]
+ExecStart=
+ExecStart=/usr/bin/cloudadapter --verbosity --debug
```
easybe, line 112: Isn‘t `-v` the default? `-vvv` is best but currently broken. `-d` is not very useful.

rettichschnidi: Yes it is, but I am betting on --debug becoming more useful once needed and -vv/-vvv being fixed soon so that I can just extend this section instead of removing it now and adding it back later on.

PR #67, opened 2026-06-23
Firmware version & update support

kelvan:
Adds firmware update support to devices: read the installed and available versions, see the update state and last update result, and trigger a pending update.
The device example CLI shows an available update inline info and gets a new update command to install it and report the outcome.

### NOTE

Set fota loop for easy testing

set `fota_loop: true` in `/etc/fwrolloutd.yml`
and restart fwrolloutd: `systemctl restart fwrolloutd`

### DIFF
--- a/gardena_smart_local_api/devices/__init__.py
+++ b/gardena_smart_local_api/devices/__init__.py
@@ -5,6 +5,7 @@
 from .device import (
     Device,
     DeviceMap,
+    FirmwareUpdateResult,
     FirmwareUpdateState,
     build_discovery_obj,
     build_inclusion_obj,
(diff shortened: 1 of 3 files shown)


easybe:
There might be an overlap with this branch: https://github.com/cloudless-garden/gardena-smart-local-api/commits/gardena/eb/fota/

kelvan:
> There might be an overlap with this branch: https://github.com/cloudless-garden/gardena-smart-local-api/commits/gardena/eb/fota/

pretty much the same it seems, just another approach to the example, let's sync tomorrow ;)

rettichschnidi (approved):
Tested via https://github.com/cloudless-garden/ha-gardena-smart-local-preview/pull/50, and works for me.

PR #68, opened 2026-06-30
Add clear/read schedules support to Gen1 irrigation devices

kelvan:
## Summary

- Allows clearing schedules on Gen1 irrigation devices via the local API, without cloud access
- Allows reading raw schedule config to validate changes (e.g. confirm schedules were cleared)

## Test plan

- [x] smart Water Control (18869) — verify `read-schedules` and `clear-schedules` work
- [x] smart Irrigation Control (31653) — verify `read-schedules` and `clear-schedules` work
- [x] smart Automatic Home & Garden Pump (22538) — verify `read-schedules` and `clear-schedules` work


rettichschnidi:
Willing to test on a IC24 classic. Any chance there is a PR for https://github.com/cloudless-garden/ha-gardena-smart-local-preview around the corner that pulls in this change?

gardena_smart_local_api/examples/irrigation.py
```diff
@@ -101,6 +102,46 @@ async def main():
                         print(f"Error: {result[0].error_message}")
                     return 1
 
+            case "clear-schedules":
+                if (wc := app.device) is None:
```
rettichschnidi, line 106: I realized it is use like this in other places already, but "wc" is quite a confusing name when ICs are covered by the same code too.

rettichschnidi (approved):
LGTM!

My setup:

```
$ uv run gardena_smart_local_api/examples/irrigation.py -g GARDENA-XXX -p XXX list
3034F8319C02C280000000BC (smart Pipeline Water Control)
3034F8319C007540000000FC (smart Irrigation Control)
3034F8EE90126D4000002D17 (smart Water Control)
3034F8EE901EE940000091A6 (Irrigation Control)
```

Works for WC:

```bash
$ uv run gardena_smart_local_api/examples/irrigation.py -g GARDENA-XXX -p XXX -d 3034F8EE90126D4000002D17 read-schedules
1 schedule(s), raw: 80200500140000
```

Also works for my IC24:

```
$ uv run gardena_smart_local_api/examples/irrigation.py -g GARDENA-XXX -p XXX -d 3034F8EE901EE940000091A6 read-schedules
3 schedule(s), raw: 807f0500140000817f0600140000827f0500140001
```

These are its schedules:

Image

Since we support reading the schedules - do we have public documentation on how to interpret the schedules?

gardena_smart_local_api/examples/irrigation.py
kelvan: true, I'll prepare a separate PR for renaming it everywhere after this one is merged

kelvan:
> Since we support reading the schedules - do we have public documentation on how to interpret the schedules?

not yet afaik, but we want to release as much docs as possible.
I have parsers for these schedules somewhere, need to clean it up and add it to the lib as an example implementation

PR #70, opened 2026-07-03
Rename wc to irrigation_device in irrigation example

kelvan:
wc misleading since example also covers irrigation control devices, not just water control.

### DIFF
--- a/gardena_smart_local_api/examples/irrigation.py
+++ b/gardena_smart_local_api/examples/irrigation.py
@@ -55,12 +55,12 @@ async def main():
                 app.list_devices()
 
             case "start":
-                if (wc := app.device) is None:
+                if (irrigation_device := app.device) is None:
                     return 1
-                assert isinstance(wc, COMPATIBLE)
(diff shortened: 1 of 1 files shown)


gardena_smart_local_api/examples/irrigation.py
```diff
@@ -55,12 +55,12 @@ async def main():
                 app.list_devices()
 
             case "start":
-                if (wc := app.device) is None:
+                if (device := app.device) is None:
```
easybe: This would be more in line with other examples:
```suggestion
                if (irrigation_device := app.device) is None:
```
Or
```suggestion
                if (is := app.device) is None:
```
for irrigation system, which we use in other applications, but, can be confusing.

easybe (approved):

gardena_smart_local_api/examples/irrigation.py
kelvan: `is` is a python keyword :)
Will change it to irrigation_device

easybe (approved):

PR #72, opened 2026-07-07
Document device type generations

rettichschnidi:
The code contains quite some abstraction based on the tech stack used on the 1st gen (Lemonbeat DDK) and 2nd gen (BNW) devices. Properly documenting the generation of each device type helps people from outside of Gardena to make sense of the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,28 +10,28 @@ Enables controlling and monitoring GARDENA smart devices in the local network, w
 
 ## Supported Devices
 
-| Device                             | Article Number                             | Model (EAN suffix) |
-| ---------------------------------- | ------------------------------------------ | ------------------ |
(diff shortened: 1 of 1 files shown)


kelvan (approved):
unsure of `1st gen` or `Gen1` would be better, as it is called Gen1/Gen2 in the code, but ok with both :)

rettichschnidi:
> unsure of `1st gen` or `Gen1` would be better, as it is called Gen1/Gen2 in the code, but ok with both :)

Gonna wait for @easybe's input then.

easybe:
Up until now we used gen1/gen2 (as a nickname) in commit messages and such. Spelling it out in documentation seem fine to me.

PR #74, opened 2026-07-14
Make example applications executable

rettichschnidi:
This allows to run all example applications without the need to in invoke the python(3) binary manually.


easybe (approved):
Thanks!

kelvan (approved):

PR #78, opened 2026-07-15
Add button config time read/write for Gen2 water controls

kelvan:
Allows reading and setting the default watering duration for the manual button on Gen2 water control / irrigation control valves, matching existing Gen1 support.

### DIFF
--- a/gardena_smart_local_api/devices/irrigation.py
+++ b/gardena_smart_local_api/devices/irrigation.py
@@ -3,9 +3,10 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 
 from abc import ABC, abstractmethod
+from typing import ClassVar, Protocol
 
 from ..messages import EgressMessageList
-from ..resources import IpsoPath
+from ..resources import VALUE_TYPES, IpsoPath
 from ._enums import _LowerNameEnum
 from .gen1 import (
     Gen1BatteryMixin,
@@ -109,6 +110,60 @@ class Gen1IrrigationScheduleMixin:
         )
 
 
+class _ButtonTimeDeviceProtocol(Protocol):
+    """Used to satisfy the type checker."""
+
+    _button_time_ipso_names: ClassVar[tuple[str, str]]
+
+    @property
+    def valve_ids(self) -> list[int]: ...
+    def get_value(self, path: IpsoPath) -> VALUE_TYPES | None: ...
+    def build_read_value_obj(self, path: IpsoPath) -> EgressMessageList: ...
+    def build_write_value_obj(
+        self, path: IpsoPath, value: VALUE_TYPES
+    ) -> EgressMessageList: ...
+    def _button_config_time_path(self, valve_id: int) -> IpsoPath: ...
+
+
+class ButtonTimeMixin:
+    """Devices with a manual button whose default watering duration is configurable."""
+
+    # (object_name, resource_name) of the button time resource.
+    _button_time_ipso_names: ClassVar[tuple[str, str]]
+
+    def _button_config_time_path(
+        self: _ButtonTimeDeviceProtocol, valve_id: int
+    ) -> IpsoPath:
+        if valve_id not in self.valve_ids:
+            raise ValueError(f"Invalid valve ID {valve_id}")
+        object_name, resource_name = self._button_time_ipso_names
+        return IpsoPath(
+            object_name=object_name,
+            object_instance_id=str(valve_id),
+            resource_name=resource_name,
+        )
+
+    def get_button_config_time(
+        self: _ButtonTimeDeviceProtocol, valve_id: int = 0
+    ) -> int | None:
+        value = self.get_value(self._button_config_time_path(valve_id))
+        if isinstance(value, int):
+            return value
+        return None
+
+    def build_refresh_button_config_time_obj(
+        self: _ButtonTimeDeviceProtocol, valve_id: int = 0
+    ) -> EgressMessageList:
+        return self.build_read_value_obj(self._button_config_time_path(valve_id))
+
+    def build_set_button_config_time_obj(
+        self: _ButtonTimeDeviceProtocol, seconds: int, valve_id: int = 0
+    ) -> EgressMessageList:
+        return self.build_write_value_obj(
+            self._button_config_time_path(valve_id), seconds
+        )
+
+
 class _Gen1Irrigation(Gen1Device, ABC):
     @property
     @abstractmethod
@@ -171,8 +226,11 @@ class Gen1WaterControl(
     Gen1IdentifyMixin,
     Gen1FrostWarningMixin,
     Gen1IrrigationScheduleMixin,
+    ButtonTimeMixin,
     _Gen1Irrigation,
 ):
+    _button_time_ipso_names = ("lemonbeat", "button_config_time")
+
     @property
     def valve_count(self) -> int:
         return 1
@@ -183,16 +241,7 @@ class Gen1WaterControl(
 
     @property
     def button_config_time(self) -> int | None:
-        value = self.get_value(
-            IpsoPath(
-                object_name="lemonbeat",
-                object_instance_id="0",
-                resource_name="button_config_time",
-            )
-        )
-        if isinstance(value, int):
-            return value
-        return None
+        return self.get_button_config_time()
 
     @property
     def temperature(self) -> int | None:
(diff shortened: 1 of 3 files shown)


gardena_smart_local_api/devices/irrigation.py
```diff
@@ -257,6 +257,47 @@ def valve_count(self) -> int:
...
+        value = self.get_value(
+            IpsoPath(
+                object_name="actuator",
+                object_instance_id=str(valve_id),
+                resource_name="default_duration_seconds",
+            )
+        )
+        if isinstance(value, int):
+            return value
+        return None
```
rettichschnidi: Untested, but the pythonic way would probably look like this:
```suggestion
        try:
            return (int)value
        except ValueError:
            return None
```

gardena_smart_local_api/devices/irrigation.py
```diff
@@ -257,6 +257,47 @@ def valve_count(self) -> int:
...
+    def get_button_config_time(self, valve_id: int = 0) -> int | None:
+        if valve_id not in self.valve_ids:
+            raise ValueError(f"Invalid valve ID {valve_id}")
+        value = self.get_value(
+            IpsoPath(
+                object_name="actuator",
+                object_instance_id=str(valve_id),
+                resource_name="default_duration_seconds",
+            )
+        )
```
rettichschnidi: Not tested, but probably more pythonic:

```suggestion
        try:
          value = self.get_value(
              IpsoPath(
                  object_name="actuator",
                  object_instance_id=str(valve_id),
                  resource_name="default_duration_seconds",
              )
          )
        except XXX:
            raise ValueError(f"Invalid valve ID {valve_id}")
```

gardena_smart_local_api/devices/irrigation.py
rettichschnidi: (same goes for other parts of the code)

gardena_smart_local_api/devices/irrigation.py
```diff
@@ -257,6 +257,47 @@ def valve_count(self) -> int:
...
+        self, seconds: int, valve_id: int = 0
+    ) -> EgressMessageList:
+        if valve_id not in self.valve_ids:
+            raise ValueError(f"Invalid valve ID {valve_id}")
+        return self.build_write_value_obj(
+            IpsoPath(
+                object_name="actuator",
+                object_instance_id=str(valve_id),
+                resource_name="default_duration_seconds",
+            ),
```
rettichschnidi: I see this code copied 4x times. How about putting it into a function?

gardena_smart_local_api/examples/irrigation.py
```diff
@@ -62,7 +70,7 @@ async def main():
                 try:
                     request = irrigation_device.build_open_valve_obj(
                         app.args.valve_id if app.args.valve_id is not None else 0,
-                        app.args.duration,
+                        app.args.duration if app.args.duration is not None else 60,
```
rettichschnidi: Would put the value 60 into a constant and use this constant for the help message too.

gardena_smart_local_api/examples/irrigation.py
```diff
@@ -150,6 +158,94 @@ async def main():
                     raw = irrigation_device.schedule_config.hex()
                     print(f"{irrigation_device.schedule_count} schedule(s), raw: {raw}")
 
+            case "read-button-time":
+                if (irrigation_device := app.device) is None:
+                    return 1
```
rettichschnidi, line 174: How about using some constants for exit codes?

gardena_smart_local_api/examples/irrigation.py
```diff
@@ -150,6 +158,94 @@ async def main():
...
+                    valve_ids = [0]
+                    request = irrigation_device.build_read_button_config_time_obj()
+                elif isinstance(
+                    irrigation_device, (Gen2WaterControl, Gen2IrrigationControl)
+                ):
+                    valve_ids = (
+                        [app.args.valve_id]
+                        if app.args.valve_id is not None
+                        else irrigation_device.valve_ids
+                    )
```
rettichschnidi: I am really not a Python pro. Maybe it is due to the (enforced) formatting. That said, I have a hard time reading (skimming) this code. How about:

```suggestion
                    # Some helpful comment here
                    if app.args.valve_id:
                        valve_ids = [app.args.valve_id]
                    else:
                        valve_ids = irrigation_device.valve_ids
```

gardena_smart_local_api/examples/irrigation.py
```diff
@@ -150,6 +158,94 @@ async def main():
...
+                    try:
+                        request = irrigation_device.build_set_button_config_time_obj(
+                            app.args.duration, valve_id
+                        )
+                    except ValueError:
+                        print(f"Invalid valve ID provided: {app.args.valve_id}")
+                        return 1
+                else:
+                    print("set-button-time not supported on this device")
+                    return 1
```
rettichschnidi: Really not sure, but I have the feeling like the complexity of deciding between different generations should not be in this function here, but handled by the overloads/mixins of the devices. Maybe like this:

```suggestion
                try:
                    request = irrigation_device.build_set_button_config_time_obj(app.args.duration, valve_id)
                except ValueError:
                    print(f"Invalid valve ID provided: {app.args.valve_id}")
                    return 1
                except NotImplemented:
                    print("set-button-time not supported on this device")
                    return 1
```

gardena_smart_local_api/examples/irrigation.py
```diff
@@ -43,9 +51,9 @@ async def main():
         {
             "name_or_flags": ["duration"],
             "nargs": "?",
-            "default": 60,
             "type": int,
-            "help": "Duration to water in seconds (default: 60s)",
+            "help": "Duration to water in seconds (start, default: 60s) or button "
+            "time in seconds (set-button-time, required)",
```
rettichschnidi: Not sure how to briefly, yet correctly to express, but the valve ID is also required.

I was confused by this not-so-helpful error message:

```console
$ uv run gardena_smart_local_api/examples/irrigation.py -g GARDENA-XXX -p XXX -d 3034F8EE90126D4000007D17 set-button-time 1300
set-button-time requires a duration (in seconds)
```

rettichschnidi (changes requested):
Some more nitty, some less nitty. Feel free to push back on whatever you see different, and I'll approve when you re-request a review.

gardena_smart_local_api/devices/irrigation.py
kelvan: you mean int(value) :)
it depends what we want to achieve here, both version do different things
getting a number string as return value is unexpected here, I would rather add some logging here to notice potential bugs.

gardena_smart_local_api/devices/irrigation.py
kelvan: get_value does not raise a exception if valve_id does not exist, it just returns None (like a .get call on a dict would)

gardena_smart_local_api/examples/irrigation.py
kelvan: good idea, I'll prepare a separate PR for this

gardena_smart_local_api/examples/irrigation.py
kelvan: better check if device has `build_set_button_config_time_obj` (if it quacks like a water control)

kelvan: or even better add a protocol that defines the feature for the devices and check if it implements the protocol

kelvan: just realized the methods are wrongly added to `_Gen2Irrigation` instead of `Gen2WaterControl`, I'll prepare a fix

gardena_smart_local_api/examples/irrigation.py
kelvan: (hopefully) improved it

gardena_smart_local_api/devices/irrigation.py
rettichschnidi: You are absolutely right - my proposal does something else. Keep yours! :=)

gardena_smart_local_api/examples/irrigation.py
```diff
@@ -60,9 +73,15 @@ async def main():
                 assert isinstance(irrigation_device, COMPATIBLE)
 
                 try:
+                    valve_id = app.args.valve_id if app.args.valve_id is not None else 0
+                    duration_seconds = (
+                        app.args.duration
+                        if app.args.duration is not None
+                        else DEFAULT_START_DURATION
+                    )
```
rettichschnidi, line 81: Does not have to be part of this PR, but I think checking if the passed argument is an integer and assigning it the default value of 60 if the user does not assign a value should be part of the argument handling code and not to be dealt with here.

rettichschnidi: Similar goes for app.args.valve_id (but probably a bit more tricky).

rettichschnidi (commented):
Prelimiar feedback. Will re-test once I have access to hardware again.

gardena_smart_local_api/examples/irrigation.py
kelvan: imo we should look into using real subcommands as the examples getting bigger, and maybe considering creating a separate cli project and reducing the examples to real easy-to-read examples :)

rettichschnidi (approved):

PR #79, opened 2026-07-15
Fix gen1 Irrigation Control name

rettichschnidi:
Just as gen2, the gen1 device was marketed as 'smart Irrigation Control'.


easybe (approved):
Never mind, I did not read well.

kelvan (approved):

PR #80, opened 2026-07-21
Declare license in project metadata, bump to 0.1.1

kelvan:
PyPI (and downstream license checks) can't detect the license without a project.license field; REUSE compliance alone doesn't populate it.


rettichschnidi (approved):
Would rather not have two independent things (version change and adding license information) in a single commit, but approving anyway.

PR #84, opened 2026-08-04
Fix Gen1 is_valve_open to detect scheduled watering

### LINKED ISSUE #83: Gen1 irrigation: is_valve_open misses schedule-driven watering (schedule_state never read)
## Observed

Device: **smart Irrigation Control, model 31653** (`protocol: gen1`), 6 valves, connected through a smart Gateway over the local WebSocket API.

On 2026-08-03 one circuit was watered **05:00–05:10** by a schedule stored in the device. The GARDENA app lists the run under "Wurde bewässert".

In Home Assistant, the corresponding `valve.*` entity stayed `closed` for the whole time — its state had not changed since the previous day, when the same valve had been opened *manually* and was reported correctly.

So: manual watering is reflected, schedule-driven watering is not.

## Analysis

`_Gen1Irrigation.is_valve_open()` in `gardena_smart_local_api/devices/irrigation.py` derives the valve state solely from the watering timer:

    def is_valve_open(self, valve_id: int = 0) -> bool | None:
        value = self.get_watering_timer(valve_id)   # lemonbeat/0/watering_timer_N
        if value is not None:
            return value > 0
        return None

`watering_timer_N` is the resource a *manual* command writes (`build_open_valve_obj` → `_build_set_watering_timer_obj`). A schedule running inside the device does not appear to set it.

The device does expose the information. From `schema/31653_irrigation_control.yaml`:

    schedule_state:
      type: vo
      access: r
      encoding: hex
      description: Schedule state (3 bytes per schedule, 108 bytes total)

`Gen1IrrigationScheduleMixin` reads `schedule_config` and offers `build_clear_schedules_obj`, but **`schedule_state` is never read** — the resource appears nowhere in the library.

By contrast, `_Gen2Irrigation.is_valve_open()` evaluates timeslot state and therefore covers scheduled runs:

    for timeslot_id in self.get_object_instance_ids("timeslot"):
        if self.get_timeslot_state(int(timeslot_id)) != TimeslotState.RUNNING:
            continue
        ...

## Suggestion

For Gen1 devices, consider deriving the valve state from `schedule_state` in addition to `watering_timer_N` — mirroring what the Gen2 implementation already does with timeslots.

Related but different: #73 covers parsing `schedule_config`; this issue is about `schedule_state`.

## Side note

`watering_timer_N` is declared as `min: -36000, max: 36000` in the schema, while `is_valve_open` tests `value > 0`. If negative values carry meaning, that check may need widening. I could not determine what a negative timer represents.

## Impact

Any automation in Home Assistant that reacts to valve state — flow monitoring, cistern level coupling, emergency shutoff — is blind to scheduled watering on Gen1 devices, which is the normal operating mode for these controllers.

## Environment

- `gardena-smart-local-api` 0.1.0
- `ha-gardena-smart-local-preview` (HACS), Home Assistant 2025.12.4
- smart Irrigation Control 31653, firmware 4.0.0-1.6.1, via smart Gateway

kelvan:
## Summary

- `watering_timer_N` read encodes priority in the sign: positive = manual, negative = scheduled, 0 = stopped
- `is_valve_open` only checked `value > 0`, so scheduled watering left the valve reported as closed
- Fix checks `value != 0` instead
- Applies to both `Gen1WaterControl` and `Gen1IrrigationControl` (both inherit `_Gen1Irrigation`)

Fixes #83

### DIFF
--- a/gardena_smart_local_api/devices/irrigation.py
+++ b/gardena_smart_local_api/devices/irrigation.py
@@ -202,7 +202,7 @@ class _Gen1Irrigation(Gen1Device, ABC):
             raise ValueError(f"Invalid valve ID {valve_id}")
         value = self.get_watering_timer(valve_id)
         if value is not None:
-            return value > 0
+            return value != 0
         return None
 
     def build_open_valve_obj(
--- a/tests/test_gen1.py
+++ b/tests/test_gen1.py
@@ -10,6 +10,7 @@ from gardena_smart_local_api.devices.gen1 import (
     Gen1Device,
 )
 from gardena_smart_local_api.devices.irrigation import Gen1WaterControl
+from gardena_smart_local_api.messages import IngressMessageList
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


easybe (approved):
I would be careful with adding "Fixes ..." to the commit/PR message and referencing the other repository as it closes issues (perhaps only in the current repo). The issue should only be closed when the lib is updated in the integration.

kelvan:
> I would be careful with adding "Fixes ..." to the commit/PR message and referencing the other repository as it closes issues (perhaps only in the current repo). The issue should only be closed when the lib is updated in the integration.

as the issue was opened in the lib project it should be resolved with the lib fix, but I guess it is also not obvious if an issue should be opened in the lib or integration (or both)

easybe:
> > I would be careful with adding "Fixes ..." to the commit/PR message and referencing the other repository as it closes issues (perhaps only in the current repo). The issue should only be closed when the lib is updated in the integration.
> 
> as the issue was opened in the lib project it should be resolved with the lib fix, but I guess it is also not obvious if an issue should be opened in the lib or integration (or both)

Ah. OK, guess I was confused then.

To reference bugs I suggest we use the `Fixes: #XY` trailer (at the bottom of the commit message, just before `Co-Authored-By:` if there is one) in the future.

Issue #82, opened 2026-08-03
Missing position data/docs about the local API


jochym:
As instructed here: https://github.com/py-smart-gardena/hass-gardena-smart-system/issues/383#issuecomment-5160245739

Are there any docs about what, and how, is avaliable on this local interface?
Is it possible to get mower position updates from the gateway?
This will be quite useful for me, and perhaps others. I have tried to develop a heat-map app monitoring how often which part of the garden is visited to help in creating better zones/start points. I was able to get some streaming of the position data as long as the map in the webapp was running but I figured it might be against the husqvarna API rules or intents to do it in that way - so I did not continue on that path. Such local api would be perfect for this purpose. Just point us to the docs.

In my particular case it is GARDENA smart SILENO life with LONA. Obviously, you need gps to get a location - I guess all LONA models will be covered.

easybe:
For Gen1 mowers like yours, you need to set `position_timer` to get positions updates. The documentation has not been published yet, but you can find it here for now: https://github.com/cloudless-garden/gardena-smart-local-api/blob/gardena/eb/icd/doc/gen1/mower_cbtl-2.6.0-dev-g55b3229_icd.html

Note that all events are also sent to the cloud, therefore, this will generate traffic on our side. If a lot of people start using this, it could become an issue. So, in the future, disabling cloud connectivity might become mandatory when using the local API.

easybe:
I implemented a first draft.

The example code can be run as follows:
```
uv run gardena_smart_local_api/examples/mower.py -g $GW -p $PW -d $DEVICE position $DURATION
```

By adding the `--dump-json` flag, all communication is written to JSON files in the current directory.

jochym:
This is one mowing session. Two hours with OpenCode. I can share the code if you wish.

Big **thank you**!

<img width="1695" height="1124" alt="Image" src="https://github.com/user-attachments/assets/1a3986f2-090c-4480-aa1a-4c6676ac3b81" />

easybe:
> This is one mowing session. Two hours with OpenCode. I can share the code if you wish.

Cool. I'll have a look for sure if you publish the code.

jochym:
This was vibe-coded with opencode ... Here is the repo:
https://github.com/jochym/gardena-mapper
I have moved to hexbin. It is running right now in my home server on docker. And here is a sample after few hours:

[...]

PR #85, opened 2026-08-04
Add mower position reporting support + tests

easybe:
As requested in #82


REUSE.toml
```diff
@@ -14,6 +14,10 @@ path = [
   "tests/data/unknown_device.json",
   "tests/data/update_event.json",
   "tests/data/water_control.json",
+  "tests/data/mower_gen1_lona.json",
+  "tests/data/mower_gen1_lona_position_update_event.json",
+  "tests/data/mower_gen2.json",
+  "tests/data/mower_gen2_position_update_event.json",
```
kelvan: Rest of the file is in alphabetical order, new entries should be ordered too

tests/test_mowers.py
```diff
@@ -0,0 +1,460 @@
...
+    assert battery is not None
+    assert isinstance(battery, float)
+    assert 0 <= battery <= 100
+    assert battery == 85
+
+
+@pytest.mark.asyncio
+async def test_mower_gen2_error(mower_gen2):
+    # Gen2Mower doesn't have an 'error' property (uses mower_app.state/activity instead)
+    pass
```
kelvan: Noop, should be removed?

tests/test_mowers.py
```diff
@@ -0,0 +1,460 @@
...
+    assert battery == 72
+
+
+@pytest.mark.asyncio
+async def test_mower_gen1_lona_state_is_charging(mower_gen1_lona):
+    # Status 8 = PARKED_WEEK_TIMER which maps to MowerState.PARKED
+    state = mower_gen1_lona.state
+    assert state is not None
+    assert isinstance(state, MowerState)
+    assert state == MowerState.PARKED
```
kelvan: Name suggests charging, assert is on parked state

tests/test_mowers.py
```diff
@@ -0,0 +1,460 @@
...
+    # Gen2Mower doesn't have an 'error' property (uses mower_app.state/activity instead)
+    pass
+
+
+@pytest.mark.asyncio
+async def test_mower_gen2_state_is_charging(mower_gen2):
+    state = mower_gen2.state
+    assert state is not None
+    assert isinstance(state, MowerState)
+    assert state == MowerState.PARKED
```
kelvan: Same, name suggests charging state, tests on parked

gardena_smart_local_api/devices/mowers.py
```diff
@@ -156,9 +164,66 @@ def build_start_mowing_obj(self, seconds: int) -> EgressMessageList:
...
+                gnss_latitude,
+                gnss_longitude,
+                gnss_horizontal_accuracy,
+                real_time_latitude,
+                real_time_longitude,
+                real_time_heading,
+                real_time_is_ready,
+                compass_heading,
+                compass_is_calibrated,
+            ) = struct.Struct(">2iI2ih?h?").unpack(data)
```
kelvan: this raises a struct error if data length is wrong, should reraise a ValueError

gardena_smart_local_api/devices/mowers.py
```diff
@@ -156,9 +164,66 @@ def build_start_mowing_obj(self, seconds: int) -> EgressMessageList:
         )
 
 
+class Gen1MowerPosition(BaseModel):
+    latitude: float
+    longitude: float
```
kelvan, line 169: not mandatory, but could a shared Position base model be beneficial?

gardena_smart_local_api/examples/mower.py
```diff
@@ -81,6 +99,32 @@ async def main():
...
+                if result is None or not result[0].success:
+                    print("Failed to start position reporting")
+                    if result is not None and isinstance(result[0], ErrorMessage):
+                        print(f"Error: {result[0].error_message}")
+                    return 1
+
+                print(f"Displaying position for {app.args.duration}s")
+                try:
+                    await asyncio.wait_for(
+                        _display_position(mower), timeout=app.args.duration
```
kelvan, line 126:
```suggestion
                        _display_position(mower), timeout=duration
```

duration is set to minimum of 60 above, but uses the raw input here

gardena_smart_local_api/examples/mower.py
```diff
@@ -81,6 +99,32 @@ async def main():
...
+                    duration = 60
+                request = mower.build_start_position_reporting_obj(duration)
+                result = await app.send_request(request)
+                if result is None or not result[0].success:
+                    print("Failed to start position reporting")
+                    if result is not None and isinstance(result[0], ErrorMessage):
+                        print(f"Error: {result[0].error_message}")
+                    return 1
+
+                print(f"Displaying position for {app.args.duration}s")
```
kelvan, line 123:
```suggestion
                print(f"Displaying position for {duration}s")
```

gardena_smart_local_api/devices/mowers.py
```diff
@@ -294,3 +442,33 @@ def build_pause_mowing_obj(self) -> EgressMessageList:
...
+        """Start position reporting for given duration.
+
+        Args:
+            seconds: Duration in seconds.
+
+        Returns:
+            EgressMessageList ready to be sent to the local GARDENA smart API.
+        """
+        INTERVAL = 3  # seconds, same value as GARDENA app
+        MIN_TIMEOUT = 60
```
kelvan, line 461: does Gen2 mowers have a stop path for the reporting?

tests/test_mowers.py
```diff
@@ -0,0 +1,460 @@
...
+
+@pytest.mark.asyncio
+async def test_mower_gen1_lona_position(mower_gen1_lona):
+    position = mower_gen1_lona.position
+    assert position is not None
+
+    assert hasattr(position, "gnss_latitude")
+    assert hasattr(position, "gnss_longitude")
+    assert position.gnss_latitude == 47.3636
+    assert position.gnss_longitude == 8.512599999999999
```
kelvan: maybe using [pytest.approx](https://docs.pytest.org/en/latest/reference/reference.html#pytest.approx) makes it cleaner

kelvan (changes requested):

kelvan:
what value does heading has if lona_is_ready is false? Does it always show north as default or does it get null (or should)

REUSE.toml
easybe: > Rest of the file is in alphabetical order

Which alphabet? :wink:

kelvan: not the cyrillic one

gardena_smart_local_api/examples/mower.py
easybe: That is correct, we only want to set the timer to 60s as we otherwise get an error.

gardena_smart_local_api/devices/mowers.py
easybe: I don't really see a benefit. I want to keep returning all the data.

gardena_smart_local_api/examples/mower.py
easybe: nope

gardena_smart_local_api/devices/mowers.py
easybe: I don't think so

easybe:
> what value does heading has if lona_is_ready is false? Does it always show north as default or does it get null (or should)

I wouldn't know.

easybe:
I let Claude rework the tests, so sorry for the large diff.

gardena_smart_local_api/examples/mower.py
kelvan: so if I call the example with 50s on a Gen2 mower it shows it runs for 50s but calls it with 60s? That does not seem right

kelvan: ok, we let the mower report for 60s even if we only show it for less than 60s?

easybe: Yes, the user wants to see position updates for 50 seconds. How else do you want to implement that? The user does not care if the position is reported longer than what they requested in the background.

gardena_smart_local_api/devices/mowers.py
easybe: We should check how the app does it.

kelvan:
> > what value does heading has if lona_is_ready is false? Does it always show north as default or does it get null (or should)
> 
> I wouldn't know.

who would know this?

gardena_smart_local_api/devices/mowers.py
kelvan: I guess this could be moved to a future PR

easybe:
> what value does heading has if lona_is_ready is false? Does it always show north as default or does it get null (or should)

I was not able to get a reading with `lona_is_ready=False`. However, this is documented in the code:

> The real time values (lat, long, heading) are used if ready, indicated with the ICD-flag RealTimeIsReady.
> If not ready the raw GNSS values (lat, long, heading) are used but only if the GNSS accuracy is good enough.
> If this is bad we do not show any position but a message that there is currently no GPS available.
>
> The compass heading is used if CompassIsCalibrated is true, otherwise always towards north.

easybe:
So, is this good to go?

kelvan:
> > what value does heading has if lona_is_ready is false? Does it always show north as default or does it get null (or should)
> 
> I was not able to get a reading with `lona_is_ready=False`. However, this is documented in the code:
> 
> > The real time values (lat, long, heading) are used if ready, indicated with the ICD-flag RealTimeIsReady.
> > If not ready the raw GNSS values (lat, long, heading) are used but only if the GNSS accuracy is good enough.
> > If this is bad we do not show any position but a message that there is currently no GPS available.
> > The compass heading is used if CompassIsCalibrated is true, otherwise always towards north.

so if it is not ready the user gets a wrong "north" value? if we do not know the heading it should be null, not north imo

easybe:
> so if it is not ready the user gets a wrong "north" value? if we do not know the heading it should be null, not north imo

done

Gen2 behaves very differently, it won't fallback to north.

kelvan (approved):

PR #89, opened 2026-08-10
Add RF link quality support for Gen2 water/irrigation controls

### LINKED ISSUE #88: Add RF link quality support for Gen2 devices
Gen1 devices expose RF link quality via `Gen1RfLinkQualityMixin` (`rf_link_quality` property + `build_refresh_rf_link_quality_obj()`), but there's no equivalent for Gen2 devices. `gen2.py` only has `Gen2BatteryMixin`, `Gen2TemperatureMixin`, and `Gen2IdentifyMixin`.

If the Gen2 protocol exposes an equivalent signal-quality value, add a matching mixin so Gen2 devices can report it too, currently they can't.

kelvan:
Gen1 devices expose RF link quality directly on `Gen1Device`, Gen2 had no equivalent. Added `radio_signal_strength` and `rf_link_quality` read from the standard LwM2M Connectivity Monitoring object, plus `build_refresh_rf_link_quality_obj()` to trigger a fresh reading (mirrors Gen1's method of the same name).

Confirmed against a real Gen2 water control (single valve) via gateway logs, values match exactly:
```
sg_common/0/measure_rf_link                       execute, triggers a measurement
connectivity_monitoring/0/radio_signal_strength    dBm
connectivity_monitoring/0/link_quality             small int scale
```

Lives directly on `Gen2Device` rather than as a per-subclass mixin: all 5 known Gen2 model schemas (water controls, irrigation controls, mowers) declare `connectivity_monitoring`, so it's universal across Gen2, not device-type-specific.

`device.py`'s generic info printer picks it up automatically too (same `getattr` pattern as `rf_link_quality`).

Only field-tested on a water control so far. @ezra, if you have a Gen2 mower handy, would be great to confirm it reports these too, otherwise we're assuming it works there based on the schema alone. Not a critical feature either way.

Closes #88

### DIFF
--- a/gardena_smart_local_api/devices/gen2.py
+++ b/gardena_smart_local_api/devices/gen2.py
@@ -72,5 +72,42 @@ class Gen2IdentifyMixin:
 
 
 class Gen2Device(Device):
+    # connectivity_monitoring is present in every known Gen2 model schema.
     model_definition: Gen2ModelDefinition = Field()
     service: ClassVar[str] = "lwm2mserver"
+
+    @property
+    def radio_signal_strength(self: _DeviceProtocol) -> int | None:
+        """Signal strength in dBm, from the standard LwM2M connectivity object.
+
+        Also refreshed by build_refresh_rf_link_quality_obj().
+        """
+        value = self.get_value(
+            IpsoPath(
+                object_name="connectivity_monitoring",
+                object_instance_id="0",
+                resource_name="radio_signal_strength",
+            )
+        )
+        return value if isinstance(value, int) else None
+
+    @property
+    def rf_link_quality(self: _DeviceProtocol) -> int | None:
+        value = self.get_value(
+            IpsoPath(
+                object_name="connectivity_monitoring",
+                object_instance_id="0",
+                resource_name="link_quality",
+            )
+        )
+        return value if isinstance(value, int) else None
+
+    def build_refresh_rf_link_quality_obj(self: _DeviceProtocol) -> EgressMessageList:
+        return self.build_execute_obj(
+            IpsoPath(
+                object_name="sg_common",
+                object_instance_id="0",
+                resource_name="measure_rf_link",
+            ),
+            None,
+        )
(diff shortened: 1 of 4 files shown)


gardena_smart_local_api/devices/gen2.py
```diff
@@ -72,5 +72,39 @@ def build_identify_obj(self: _DeviceProtocol) -> EgressMessageList:
 
 
 class Gen2Device(Device):
+    # connectivity_monitoring is present in every known Gen2 model schema.
     model_definition: Gen2ModelDefinition = Field()
     service: ClassVar[str] = "lwm2mserver"
+
+    @property
+    def radio_signal_strength(self: _DeviceProtocol) -> int | None:
+        """Signal strength in dBm, from the standard LWM2M connectivity object."""
```
rettichschnidi:
```suggestion
        """Signal strength in dBm, from the standard LwM2M connectivity object."""
```

gardena_smart_local_api/devices/gen2.py
```diff
@@ -72,5 +72,39 @@ def build_identify_obj(self: _DeviceProtocol) -> EgressMessageList:
...
+        value = self.get_value(
+            IpsoPath(
+                object_name="connectivity_monitoring",
+                object_instance_id="0",
+                resource_name="link_quality",
+            )
+        )
+        return value if isinstance(value, int) else None
+
+    def build_measure_rf_link_obj(self: _DeviceProtocol) -> EgressMessageList:
```
easybe:
```suggestion
    def build_refresh_rf_link_quality_obj(self: _DeviceProtocol) -> EgressMessageList:
```
to align with Gen1

easybe: Maybe add comment that signal strength is also updated.

easybe (changes requested):

gardena_smart_local_api/devices/gen2.py
easybe: The commit message also misspells LwM2M.

easybe (approved):

PR #92, opened 2026-08-24
Fix: report flow_rate=0 when pump is not running

cy-bertrand:
The gateway reports a default/calibration value of 100 L/h for flow_rate even when the pump is stopped. This mirrors the fix i've already applied in the Home Assistant integration (cloudless-garden/ha-gardena-smart-local-preview#113) by using the existing is_running property to short-circuit and return 0 when the pump is confirmed off (while still returning None if the running state hasn't been read yet.)

Should resolve the issue https://github.com/cloudless-garden/gardena-smart-local-api/issues/93

### DIFF
--- a/gardena_smart_local_api/devices/irrigation.py
+++ b/gardena_smart_local_api/devices/irrigation.py
@@ -521,6 +521,11 @@ class Pump(
 
     @property
     def flow_rate(self) -> int | None:
+        # The device can report values from 100 to 13615. Therefore, even when the pump
+        # is not running, a value of 100 gets reported.
+        # The following condition avoids passing on obviously wrong values.
+        if self.is_running is False:
+            return 0
         value = self.get_value(
             IpsoPath(
                 object_name="lemonbeat",
--- a/tests/test_epp.py
+++ b/tests/test_epp.py
@@ -63,7 +63,7 @@ async def test_epp_outlet_temperature(epp):
 
 @pytest.mark.asyncio
 async def test_epp_flow_rate(epp):
-    assert epp.flow_rate == 100
+    assert epp.flow_rate == 0
 
 
 @pytest.mark.asyncio


gardena_smart_local_api/devices/irrigation.py
```diff
@@ -518,9 +518,14 @@ def outlet_temperature_min(self) -> int | None:
         if isinstance(value, int):
             return value
         return None
-
+
```
rettichschnidi: Unrelated change, please undo.

gardena_smart_local_api/devices/irrigation.py
```diff
@@ -518,9 +518,14 @@ def outlet_temperature_min(self) -> int | None:
         if isinstance(value, int):
             return value
         return None
-
+        
     @property
     def flow_rate(self) -> int | None:
+    # The gateway reports flow_rate=100 as a default/calibration value
+    # even when the pump is stopped. Only report a real flow when the
+    # pump is actually running.
```
rettichschnidi: I think this is a bit more useful:

```suggestion
    # The device can report values from 100 to 13615. Therefore, even when the pump
    # is not running, a value of at least 100 gets reported.
    # The following condition avoids passing on obviously wrong values.
```

rettichschnidi (changes requested):

gardena_smart_local_api/devices/irrigation.py
cy-bertrand: Ok.
Actually not at "least 100" but exactly 100 😜

rettichschnidi (changes requested):
Thanks!

Now please merge those two commits into one and make sure to follow the gitlint findings (CI has some complains).

If you do not know how to do this let us know. We'll either explain it or will do it for you.

easybe:
> Thanks!
> 
> Now please merge those two commits into one and make sure to follow the gitlint findings (CI has some complains).
> 
> If you do not know how to do this let us know. We'll either explain it or will do it for you.

Some hints are given here: https://github.com/cloudless-garden/gardena-smart-local-api#commits

gardena_smart_local_api/devices/irrigation.py
```diff
@@ -521,6 +521,11 @@ def outlet_temperature_min(self) -> int | None:
 
     @property
     def flow_rate(self) -> int | None:
+    # The device can report values from 100 to 13615. Therefore, even when the pump
+    # is not running, a value of 100 gets reported.
+    # The following condition avoids passing on obviously wrong values.
+        if self.is_running is False:
```
easybe, line 527:
```suggestion
        if not self.is_running:
```

gardena_smart_local_api/devices/irrigation.py
```diff
@@ -521,6 +521,11 @@ def outlet_temperature_min(self) -> int | None:
 
     @property
     def flow_rate(self) -> int | None:
+    # The device can report values from 100 to 13615. Therefore, even when the pump
+    # is not running, a value of 100 gets reported.
+    # The following condition avoids passing on obviously wrong values.
```
easybe:
```suggestion
        # The device can report values from 100 to 13615. Therefore, even when the pump
        # is not running, a value of 100 gets reported.
        # The following condition avoids passing on obviously wrong values.
```

easybe (changes requested):

gardena_smart_local_api/devices/irrigation.py
cy-bertrand: Hi @easybe,

If I'm not wrong if we put "if not self.is_running:" then the value will report 0 instead of "none" when the device status is not know. Am I'm wrong ?

I'd rather not assert a fact (zero flow) we haven't actually verified. So I'll keep is False here — let me know if you see a case where that causes trouble.

easybe: Yes, you are right, did not think of that.

cy-bertrand:
ok i've tried to squash the commits into one and change the text to be gitlints compliant

rettichschnidi (changes requested):
Would delete the following in the Git commit, because it is not relevant for the history of this repository and that PR will not be merged:

> This mirrors the fix already applied in the Home Assistant integration (https://github.com/cloudless-garden/ha-gardena-smart-local-preview/pull/113) by using the existing is_running property to short-circuit and return 0 when the pump is confirmed off (while still returning None if the running state hasn't been read yet). 

Everything else LGTM.

rettichschnidi (approved):

easybe:
Tests need to be adapted too...

cy-bertrand:
> Tests need to be adapted too...

**Suggestion for the test**

Since the new logic overrides `flow_rate` to `0` when the pump is stopped, the test should reflect that behavior. The fixture can stay as-is (with the device reporting `flow_rate: 100` while `pump_state: 0` — a realistic edge case), and we just update the expected result:

In `test_epp.py`, change:

```python
assert epp.flow_rate == 100
```

to:

```python
assert epp.flow_rate == 0
```

This validates the fix: even when the underlying device reports `100`, the computed `flow_rate` correctly returns `0` because the pump is idle.

rettichschnidi:
@cy-bertrand Rather than suggesting fixes, you should implement them and get CI to pass. Without, the pipeline fails and we can not merge this PR.

kelvan (approved):
lgtm

easybe (approved):

PR #96, opened 2026-09-18
Add a star history chart to the README

kelvan:
## Summary

Embeds a [star-history.com](https://star-history.com) chart in the README, light/dark aware.

I know it's not the most important feature 😅

PR #98, opened 2026-09-18
Revert "Add a star history chart to the README"

kelvan:
Reverts #96.


easybe (approved):
So far it is 2 👍 votes against my 👎, so no need for a revert, but I’ll take it.

